Issue #385, opened 2019-02-05
Make sure that UI works on mobile devices


Marcel-Ebert:
## Left to do (2019-04-02)

- [x] MainTitle wrapping & font-size (#468)
- [ ] Account page
  - [ ] Balances in one line
- [x] Send payment: Use `<MainTitle>`
- [x] Tx confirmation dialog:
  - [ ] Full-screen
  - [ ] Use `<MainTitle>`
  - [ ] Font-sizes
- [ ] ~~Move main action buttons to bottom & make sticky on mobile~~ (moved to #467)

Marcel-Ebert:
Transactions appearing in the SignatureRequestList can not be reviewed or dismissed.

Probably because these options only appear when hovering above them, which is not possible on mobile(?)

Marcel-Ebert:
In "Receive Funds" dialog copying the text to clipboard by clicking on it does not work.
`Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'navigator.clipboard.writeText')
https://stellar-horizon-testnet.satoshipay.io/accounts/GDESJFVIMGWZQQI33YPTGKZ6PBNTNFY5FVVLTVBGMSK3TTQMPT3CQOZJ?c=0.8019324937122966`

The hint could be removed on mobile versions because the user is able to long press on the text and copy it by himself or the clipboard can be accessed via [Cordova Clipboard Plugin](https://www.npmjs.com/package/cordova-clipboard).

andywer:
> Transactions appearing in the SignatureRequestList can not be reviewed or dismissed.

Can't you tap on the item to trigger the hover state? 🤔

> copying the text to clipboard by clicking on it does not work

Good point! Let's create a `clipboard.ts` file in `src/platform`, shall we?

Marcel-Ebert:
> Can't you tap on the item to trigger the hover state? 

It's strange. It does not seem to work at all on iOS Simulator but it worked sometimes on a physical android device although it was kind of buggy. If you click on the right side of the element you instantly select the option beneath it before seeing it and afterwards clicking on it seems to be disabled for some time. 

> Let's create a clipboard.ts file in src/platform, shall we?

Sure, let's use the Cordova Clipboard Plugin then.

Marcel-Ebert:
Space between ListItem text and balance might be too big (?)

[...]

PR #468, opened 2019-04-02
Improve MainTitle & account page styles (mobile)

andywer:
Related to #385.


andywer:
@Marcel-Ebert Can be reviewed now :)

PR #472, opened 2019-04-04
Use <MainTitle> in Send-Payment Dialog

Marcel-Ebert:
Related to #385.

We should think about editing the <MainTitle> component so that the position of the badges prop can be determined by the component using it.

### DIFF
--- a/src/components/Payment/CreatePaymentDialog.tsx
+++ b/src/components/Payment/CreatePaymentDialog.tsx
@@ -88,7 +88,6 @@ function CreatePaymentDialog(props: Props) {
         </Box>
         <CreatePaymentForm
           accountData={props.accountData}
-          onCancel={props.onClose}
           onSubmit={handleSubmit}
           trustedAssets={trustedAssets}
           txCreationPending={txCreationPending}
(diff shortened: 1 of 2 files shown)


src/components/Payment/CreatePaymentDialog.tsx
```diff
@@ -79,15 +79,15 @@ function CreatePaymentDialog(props: Props) {
   return (
     <Dialog open={props.open} fullScreen onClose={props.onClose} TransitionComponent={Transition}>
       <Box width="100%" maxWidth={900} padding="24px 36px" margin="0 auto">
-        <Typography variant="h5" component="h2" style={{ marginTop: 8, marginBottom: 8 }}>
-          Send funds {props.account.testnet ? <TestnetBadge style={{ marginLeft: 8 }} /> : null}
-        </Typography>
+        <MainTitle
+          title={<span>Send funds {props.account.testnet ? <TestnetBadge style={{ marginLeft: 8 }} /> : null}</span>}
```
Marcel-Ebert: I added the badge component directly to the title prop so that it is always next to the title instead of floating on the right side if the device width is too small.

PR #511, opened 2019-04-10
Replace sdkversion numbers in build.gradle with required values

Marcel-Ebert:
- [x] Add gradle arguments to android signing script 
- [x] Make `patch-android-config.sh` insert information about `minSdkVersion` and `targetSdkVersion ` to app/build.gradle + CordovaLib/build.gradle and adjust app-`AndroidManifest` to override settings settings from CordovaLib on conflict


Marcel-Ebert:
I will look into this once again as it seems that it still does not completely fix the signing issue.

Marcel-Ebert:
Seems like a fragile solution but at least it's working now as expected, meaning that a signed build with the correct values will be created and the project still works in Android Studio.

PR #548, opened 2019-04-18
Use WKWebview / Fix pasting issue on iOS

### LINKED ISSUE #515: Cannot paste into textfields on iOS / WkWebView
~~Potential fix: <https://stackoverflow.com/questions/27188234/cant-copy-paste-in-phonegap-ionic-ios>~~

~~We might have to set up a clipboard paste handler and write the value into the `document.activeElement` manually...~~

Might be a bug with UIWebView which they are not gonna fix anymore. Might need to switch to WkWebView.

Marcel-Ebert:
- [x] Add `cordova-plugin-wkwebview-engine` to make use of other webview which resolves pasting issue on ios
- [x] Add `cordova-plugin-keyboard` to get access to keyboard events on ios
- [x] Add `cordova-plugin-device` to get access to device information
- [x] Add method to `app.cordova.ts` which resolves issue of shifted viewport on iphone after keyboard dismissal

Closes #515.

### DIFF
--- a/cordova/config/config.template
+++ b/cordova/config/config.template
@@ -81,6 +81,12 @@
         <edit-config target="NSCameraUsageDescription" file="*-Info.plist" mode="merge">
             <string>The app needs access to the camera, to be able to read QR codes.</string>
         </edit-config>
+
+        <feature name="CDVWKWebViewEngine">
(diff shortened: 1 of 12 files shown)


andywer:
Still can't paste on my iPhone 6 😕

PR #565, opened 2019-04-25
Disable text selection in some components

### LINKED ISSUE #469: Disable text selection in some components
Use the `user-select`-CSS-property to prevent the text selection in components where selecting/copying the current text should not be desired/possible.

Components where text should stay selectable:
- public keys 
- stellar addresses

Marcel-Ebert:
- [x] Add css styling to disable `user-select` everywhere except for `input` and `textarea`
- [x] Add `user-select: "text"` attribute to `<PublicKey>` to re-enable text selection for public keys.

Closes #469.

Do we need to add an exception for `<Address>` components as well? I was not 100% sure, because they contain a `<PublicKey>` as well. (@andywer)

### DIFF
--- a/src/base-styles.css
+++ b/src/base-styles.css
@@ -16,23 +16,31 @@ body {
 }
 
 .iphone-notch-top-spacing {
-  padding-top: constant(safe-area-inset-top); /* iOS 11.0 */
-  padding-top: env(safe-area-inset-top); /* iOS 11.2 */
+  padding-top: constant(safe-area-inset-top);
+  /* iOS 11.0 */
+  padding-top: env(safe-area-inset-top);
+  /* iOS 11.2 */
 }
 
 .iphone-notch-left-spacing {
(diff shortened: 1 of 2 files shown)


andywer:
> Do we need to add an exception for <Address> components as well? I was not 100% sure, because they contain a <PublicKey> as well.

Yes, `<Address>`, too :)

andywer:
TODO: Test it (@andywer)

src/base-styles.css
```diff
@@ -16,23 +16,31 @@ body {
 }
 
 .iphone-notch-top-spacing {
-  padding-top: constant(safe-area-inset-top); /* iOS 11.0 */
-  padding-top: env(safe-area-inset-top); /* iOS 11.2 */
+  padding-top: constant(safe-area-inset-top);
```
andywer: What's the reason behind those changes? Looks like that was not intended 😉

Marcel-Ebert: My prettier extension moved them to a new line, don't ask me why

PR #568, opened 2019-04-26
Request fingerprint/FaceID to unlock on mobile devices

### LINKED ISSUE #454: iOS/Android: Request fingerprint to unlock
Make the user unlock the app with their fingerprint or Face ID first. This might be the right package: [`cordova-plugin-fingerprint-aio`](https://www.npmjs.com/package/cordova-plugin-fingerprint-aio)

Demand fingerprint auth on:
- Application launch
- When re-opening the app after leaving it

Marcel-Ebert:
- [x] Add `cordova-plugin-fingerprint-aio` for fingerprint and FaceID support
- [x] Add `SplashScreenHandler` which can show/hide the SplashScreen based on message events from mainframe  
- [x] Add a `clientSecret` to the cordova securestorage (required for the android implementation of the fingerprint authentication)

On application launch and reenter, a native dialog will appear which requests the user defined fingerprint to continue. If the dialog is canceled the application will close.

Closes #454.


Marcel-Ebert:
There is a bug on first app start on android. The dialog will show invalid user input although no user action was involved, I am still investigating that issue.

Marcel-Ebert:
Fixed the bug I mentioned before. The dialog hung up because the authentication was started in `onDeviceReady` and `onResume` so twice in a row. That's why I added a check if `startAuthentication` was already called in the last 5 seconds.

src/cordova/storage.ts
```diff
@@ -106,7 +107,8 @@ async function prepareStorage(secureStorage: CordovaSecureStorage) {
   await Promise.all([
     addPlaceholderKey(storeKeys.keystore, {}),
     addPlaceholderKey(storeKeys.settings, {}),
-    addPlaceholderKey(storeKeys.ignoredSignatureRequests, [])
+    addPlaceholderKey(storeKeys.ignoredSignatureRequests, []),
+    addPlaceholderKey(storeKeys.clientSecret, "TODO_RANDOM_NUMBER")
```
Marcel-Ebert: We still have to add the generation of the random number here but I wasn't sure how to do this best.

src/cordova/app.cordova.ts
```diff
@@ -31,6 +35,58 @@ function onDeviceReady() {
...
+}
+
+function showDialogue(clientSecret: string) {
+  showAuthenticationDialogue(clientSecret)
+    .then(hideSplashScreen)
+    .catch(exitApp)
+}
+
+function exitApp() {
+  window.navigator.app.exitApp()
```
Marcel-Ebert: The `app` type is not available on `window.navigator` so we have to define it somehow somewhere.

andywer:
TODO: Test on actual iOS device.

andywer:
@Marcel-Ebert Please review my latest commit (<https://github.com/satoshipay/solar/pull/568/commits/8dbcefcb5e15e78f04c0eb9e9ce867ee4523b63d>) and test again on Android 🤖

Marcel-Ebert:
I currently encounter [this issue](https://github.com/NiklasMerz/cordova-plugin-fingerprint-aio/issues/83) when testing on my physical android device (running Android 9). We need to figure out a "simple" solution/workaround for this because right now it hinders the app from starting.

PR #576, opened 2019-05-09
Put fingerprint lock behind feature flag

andywer:
Default is off.

@Marcel-Ebert Can you please try on Android? I don't expect any surprises there, but just to be safe 😉

### DIFF
--- a/src/context/settings.tsx
+++ b/src/context/settings.tsx
@@ -14,18 +14,21 @@ interface Props {
 
 interface ContextType {
   agreedToTermsAt: string | undefined
+  biometricLock: boolean
   confirmToC: () => void
   ignoreSignatureRequest: (signatureRequestHash: string) => void
   ignoredSignatureRequests: string[]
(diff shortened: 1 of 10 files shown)


Marcel-Ebert:
I tested it on android and it does work overall, BUT I found a scenario where the application stops working:
Launch the app without fingerprint lock enabled, go to settings, turn on fingerprint lock, leave the application (press home), then reenter the application and it does not respond anymore. 
I tried capturing log statements but no error messages are shown.

@andywer maybe you have any idea why this could happen? It does not occur the other way around and it will work as expected when completely shutting down the app.

andywer:
@Marcel-Ebert i think i know why! in `app.cordova.ts` the client secret promise will not have been set yet...

andywer:
@Marcel-Ebert Can you try again, pls?

Marcel-Ebert:
Tested it on physical device and emulator and is working as expected now. 👍

PR #592, opened 2019-05-16
Set color for android statusbar

### LINKED ISSUE #570: Set color for statusbar on Android
On Android devices the application currently has a black status bar (the bar on the very top which shows time, notifications, etc...). 
We might want to change this to a blue color because first, it does not look good and second, differs from what it looks like on iOS.

To do that we could use the [cordova-plugin-statusbar](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-statusbar/).

Marcel-Ebert:
- [x] Add 'cordova-plugin-statusbar'
- [x] Set custom color for android statusbar on app start in `app.cordova.ts`
- [x] Change the color of the text (clock and battery) in the ios statusbar from black to white  

~~Note: I replaced `process.env.PLATFORM` with `cordova.platformId` because the environment variable is undefined when running the `dev`-builds (`npm run dev:android` and `npm run dev:ios`). It is however set when running `npm run build:android` and `npm run build:ios`. But to make it work in both variants I decided to change it this way.~~

Closes #570.

### DIFF
--- a/src/cordova/app.cordova.ts
+++ b/src/cordova/app.cordova.ts
@@ -10,7 +10,9 @@ import { getCurrentSettings, initSecureStorage, storeKeys } from "./storage"
 import { bioAuthenticate, isBiometricAuthAvailable } from "./bio-auth"
 
 const iframe = document.getElementById("walletframe") as HTMLIFrameElement
-const showSplashScreenOnIOS = () => (process.env.PLATFORM === "ios" ? navigator.splashscreen.show() : undefined)
+const showSplashScreenOnIOS = () => (cordova.platformId === "ios" ? navigator.splashscreen.show() : undefined)
+const setStatusbarColor = () =>
+  cordova.platformId === "android" ? StatusBar.backgroundColorByHexString("#D601a4ed") : undefined
 
 let bioAuthInProgress: Promise<void> | undefined
 let bioAuthAvailablePromise: Promise<boolean>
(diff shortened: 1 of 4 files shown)


src/cordova/app.cordova.ts
```diff
@@ -10,7 +10,9 @@ import { getCurrentSettings, initSecureStorage, storeKeys } from "./storage"
 import { bioAuthenticate, isBiometricAuthAvailable } from "./bio-auth"
 
 const iframe = document.getElementById("walletframe") as HTMLIFrameElement
-const showSplashScreenOnIOS = () => (process.env.PLATFORM === "ios" ? navigator.splashscreen.show() : undefined)
+const showSplashScreenOnIOS = () => (cordova.platformId === "ios" ? navigator.splashscreen.show() : undefined)
```
andywer: Let's better add `PLATFORM=<platform>` to the `dev` npm scripts in `cordova/package.json`. Just for consistency's sake, so we don't need to think about when to use which one ;)

src/cordova/app.cordova.ts
```diff
@@ -10,7 +10,9 @@ import { getCurrentSettings, initSecureStorage, storeKeys } from "./storage"
...
+const setStatusbarColor = () =>
+  cordova.platformId === "android" ? StatusBar.backgroundColorByHexString("#D601a4ed") : undefined
```
andywer: Need to add `StatusBar` to the `cordova.d.ts` 😉

src/cordova/app.cordova.ts
Marcel-Ebert: It actually is defined in the scripts, I looked it up before changing it this way but it does not work anyway.

I checked: If you run `dev:android` or `dev:ios` then `process.env.PLATFORM` will be defined within the application code but not within our cordova code. 
I could not come up with a solution for this maybe you have an idea what could be the reason behind this?

PR #593, opened 2019-05-16
Show numeric keyboard for number input fields

### LINKED ISSUE #580: Automatically show a numeric keyboard for number input fields
Show a numeric keyboard instead of the default one when the user clicks on input fields which expect numeric input.

Marcel-Ebert:
- [x] Add `type="number"` to the `<TextField>`s that should only accept numeric input
- [x] Hide the up/down spin buttons which are shown by default for input fields with type `number`

Closes #580.

### DIFF
--- a/src/base-styles.css
+++ b/src/base-styles.css
@@ -54,3 +54,9 @@ body {
   user-select: none;
   -webkit-user-select: none;
 }
+
+input[type="number"]::-webkit-inner-spin-button,
+input[type="number"]::-webkit-outer-spin-button {
+  -webkit-appearance: none;
+  margin: 0;
+}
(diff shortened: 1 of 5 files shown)


src/base-styles.css
```diff
@@ -54,3 +54,9 @@ body {
   user-select: none;
   -webkit-user-select: none;
 }
+
+input[type="number"]::-webkit-inner-spin-button,
+input[type="number"]::-webkit-outer-spin-button {
+  -webkit-appearance: none;
```
andywer: What is the `-webkit-appearance: none` good for?

Marcel-Ebert: Apparently it can be used to hide the buttons. I used it because it is mentioned in the examples on how to hide them and is supported on Chrome and Safari.

PR #597, opened 2019-05-22
Show ConfirmDialog on mainnet account creation without password

### LINKED ISSUE #260: Warn if creating mainnet account without password
Warn the user that this is an unsafe action and that anyone accessing the computer will be able to access their account funds. Show the warning as a confirmation dialog.

Marcel-Ebert:
Show a confirmation dialog when the user attempts to create a mainnet account without password protection. 

Closes #260.


src/components/Form/CreateAccount.tsx
```diff
@@ -72,6 +72,8 @@ interface AccountCreationFormProps {
 
 function AccountCreationForm(props: AccountCreationFormProps) {
   const { errors, formValues, setFormValue } = props
+  const [pendingConfirmation, setPendingConfirmation] = React.useState<React.SyntheticEvent | null>(null)
```
andywer: Super subtle issue here: We cannot actually store the event here, since we will only use the event to do a `.preventDefault()` in the parent component.

But: `preventDefault()` only works when called synchronously.

Solution: Let's call `event.preventDefault()` in this component now that it sets up its own `onSubmit` handler anyway and not expose the event in its `props.onSubmit()`.

PR #600, opened 2019-05-24
Replace account renaming dialog

### LINKED ISSUE #246: Replace account renaming dialog
Turn the account name into a text field instead. Still triggered via menu item.

Alternatively re-use "Create account" page. Check with Daniel.

Marcel-Ebert:
- Add `<InlineEditField>` component
- Use the `<InlineEditField>` instead of the `<Typography>` component in `<MainTitle>`
- Change the flow of the account renaming process so that on select of "Rename" in the account context menu no dialog opens but the `<InlineEditField>` will switch to edit mode and change the account name after the user has finished editing (by hitting "Enter" or clicking somewhere else)

Closes #246.


andywer:
Superseded by #663.

PR #602, opened 2019-05-29
Add password-strength indicator

Marcel-Ebert:
- [x] Add `zxcvbn` dependency for calculation of password strength
- [x] Create `<PasswordStrengthTextField>` component
- [x] Use `<PasswordStrengthTextField>` in account creation form and in change-password dialog


andywer:
Closing for now. Let's look into this change set again once we start working on the app-wide password.

PR #614, opened 2019-06-06
Show tx hash and link to stellar.expert

andywer:


andywer:
No matter if we ship the external inspection link button or not, but we should definitely show the tx hash when opening past txs.

PR #620, opened 2019-06-13
Implement POC protocol handler

Marcel-Ebert:
- [x] Add protocol handler for electron platforms
- [x] Add protocol handler for cordova platforms
- [x] Register application as client for `web+stellar` requests
- [x] Rename linux build artifact so that it does not contain whitespaces (required for the resulting `.desktop` file on install)
- [x] Add method to subscribe to messages from `ipcMain`
- [x] Adjust android config patch file
- [x] Set `AndroidLaunchMode` to `singleTask` so that if the application through a link the existing application will be reused instead of a second instance being created

### DIFF
--- a/src/platform/electron/protocol-handler.ts
+++ b/src/platform/electron/protocol-handler.ts
@@ -1,0 +1,11 @@
+export function subscribeToDeepLinkURLs(callback: (url: string) => void) {
+  if (!window.electron) {
+    throw new Error("No electron runtime context available.")
+  }
+
+  return window.electron.subscribeToIPCMain("deeplink:url", (event, args) => {
+    if (args) {
+      callback(args)
+    }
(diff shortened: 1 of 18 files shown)


src/platform/electron/protocol-handler.ts
```diff
@@ -0,0 +1,11 @@
+export function subscribeToDeepLinkURLs(callback: (url: string) => void) {
+  if (!window.electron) {
+    throw new Error("No electron runtime context available.")
+  }
+
+  return window.electron.subscribeToIPCMain("deeplink:url", (event, args) => {
```
andywer: Let’s rename `args` to `url`. Btw, regarding the line below: Is there a real risk that the IPC message body (`args`) is not set?

PR #652, opened 2019-07-11
Add inflation management to account settings

### LINKED ISSUE #287: Inflation management
Allow setting an inflation destination (see Stellar docs) as part of the account settings.

Low-hanging fruits: Just show our usual destination address (public key of inflation pool address) input field, no one-click options.

Marcel-Ebert:
- [x] Create `<InflationDestination>` dialog
- [x] Add "Inflation Destination" menu item to context menu in `AccountHeaderCard`
- [x] Show different view for `SetOptions` operation in `TransactionReview` if the operation contains an `inflationDest`

Closes #287.

### DIFF
--- a/src/components/AccountSettings/AccountSettings.tsx
+++ b/src/components/AccountSettings/AccountSettings.tsx
@@ -6,6 +6,7 @@ import Slide from "@material-ui/core/Slide"
 import { TransitionProps } from "@material-ui/core/transitions/transition"
 import EyeIcon from "@material-ui/icons/RemoveRedEye"
 import DeleteIcon from "@material-ui/icons/Delete"
+import DestinationIcon from "@material-ui/icons/Place"
 import GroupIcon from "@material-ui/icons/Group"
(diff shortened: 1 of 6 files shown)


andywer:
Looks like SDF is not gonna get rid of inflation too soon.

Two things left to do here:
- Let's improve the UX a bit
- Wait for #663 to get merged

## UX improvement

Since the story was very poorly spec-ed before, here an idea how to make the feature more convenient:

Show the inflation destination read-only first. If the public key can be resolved to a domain, show the domain.

Show an edit button. When the users clicks it, switch the field from read-only to editable. Don't show the resolved domain anymore, only use the public key in edit mode. Show a "save" and "cancel" button in edit mode.

andywer:
Rebased on latest `master`. The comment above is still valid.

andywer:
We might or might not provide a one-click list of popular inflation pools, but it's safe to merge this PR now.

PR #654, opened 2019-07-12
Prevent sending funds to exchanges without memo

### LINKED ISSUE #613: Prevent sending funds to exchanges without memo
New wallet guideline:
When making a payment, check whether destination is a well-known account of an exchange. If so, prevent creating a tx without memo ~~or show a warning and ask for confirmation~~.

List of exchange accounts can be hardcoded somewhere central in the code base.

See also:
- https://stellar.expert/directory/public/
- https://github.com/orbitlens/stellar-expert-explorer/blob/master/docs/api/directory.md

Marcel-Ebert:
**NOT READY** for merge yet, blocked by #735.

- [x] Add `well-known-stellar-accounts.json`-file containing a list of well-known stellar accounts from [Stellar Expert](https://stellar.expert/directory/public/).
- [x] Add function that checks if the data contains an account matching the passed `publicKey` and if it is tagged as 'exchange'.
- [x] Check if the selected `destination` is the address of a known exchange before creating a payment. If it is, change the placeholder text of the memo textfield from "Description (optional)" to "Description (mandatory)"  and show an error label if the user tries to send the payment without a memo. 

Closes #613.


Marcel-Ebert:
The JSON file was modified so that it only contains information about exchanges instead of all well-known stellar accounts from StellarExpert.

andywer:
After careful (re-)consideration: We should fetch the stellar.expert data dynamically and cache it locally. This data will be valuable for other use cases as well and that way we don't have to manually keep it up-to-date.

See #732.

Marcel-Ebert:
Superseded by #753.

PR #662, opened 2019-07-19
Migrate to Material-UI v4

Marcel-Ebert:
- [x] Update material-ui dependencies
- [x] Update storybook dependencies (older versions wouldn't compile)
- [x] Replace `unstable_useMediaQuery` usages
- [x] Wrap transitions with `React.forwardRef`
- [x] Change style definition in `<NotificationContainer>`
- [x] Replace usages of `"default"` in typography colours with `"initial"`
- [x] Add `'display: "block"` as typography prop to `<ListItemText>` components (because `'display: "block"` is not used as default style anymore)


Marcel-Ebert:
@andywer we could set the `align:"right"` prop for `MuiListItemText` in our theme but I chose not to because it does not apply to all our `<ListItemText>` components so it seems to be easier to leave them as it is for now.

andywer:
Rebased on latest `master`.

PR #647, opened 2019-07-05
Support StellarGuard setup via SEP-7 request

ebma:
- [x] Create `transactionRequest` context which receives the `web+stellar` request and parses+verifies URIs
- [x] Create `TransactionRequestHandler` which renders a component based on the Transaction URI
- [x] Create `StellarGuardActivationDialog` 
- [x] Add `stellar-uri` dependency for URI parsing (+ `babel-polyfill` dependency because `stellar-uri` doesn't compile without it)
- [x] Add button to create a new account to `AccountSelectionList` component
- [x] Add function that creates a copy of a `Transaction` but replaces the source account to `lib/transaction.ts` (needed in `StellarGuardActivation` to replace the source account from the transaction contained in the SEP-0007 transaction with the account that was selected by the user)
- [x] Add `account` to `sendTransaction()` signature  and remove the `account` prop from `TransactionSender` to support sending transactions in cases where the sending account is unknown at first

PR #702, opened 2019-08-21
Create handler for secret key import

Marcel-Ebert:
**This branch is based on the StellarGuard branch (see #647).**

- [x] Add `SolarUri` class
- [x] Add optional `importSecretKey` prop to `CreateAccount` form and if this property is present disable the switch and textfield for importing an account and fill the textfield with the value of the `importSecretKey` prop so that the user has to import the specified secret key
- [x] Make transactionRequest context parse a `StellarUri` or `SolarUri` depending on the pathname of the incoming uri
- [x] When a `SolarUri` was passed to the protocol handler: show the modified `CreateAccount` page with the secretKey that was extracted from the uri


andywer:
We won't be able to do it like this, but there is an alternative. Closing.

PR #721, opened 2019-08-30
Remove window menu of Windows app

andywer:
Remove the ugly "File", "Edit", ... menu at the top of the window.


andywer:
Was ok in v0.20.1. Must have slipped in since then (electron update?).

PR #735, opened 2019-09-05
Create hook for fetching of stellar.expert directory data

### LINKED ISSUE #732: Fetch stellar.expert directory data
On every launch of the app fetch the stellar.expert account directory data via their open API (always or if it hasn't been fetched in 24h or so).

Cache it locally (just put it in local storage), so the data is immediately available. See `useStellarTomls()` hooks for a similar implementation.

Marcel-Ebert:
- [x] Add type definition for the stellar.expert data records
- [x] Create a `useWellKnownAccounts()` hook 

This hook checks the local-storage for cached data and if this data is unavailable or was downloaded more than 24h ago, the stellar.expert directory api will be fetched and new local-storage entries created based on the response.
The hook provides a `lookup` method that takes a publicKey and returns the corresponding `AccountRecord` (if any).

Closes #732 .

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -33,6 +33,7 @@ import {
   ObservedTradingPair
 } from "./subscriptions"
 import * as Clipboard from "./platform/clipboard"
+import { AccountRecord } from "./types/well-known-accounts"
 import { StellarToml, StellarTomlCurrency } from "./types/stellar-toml"
 
 export { ObservedAccountData, ObservedRecentTxs, ObservedTradingPair }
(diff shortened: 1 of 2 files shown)


Marcel-Ebert:
OK I think I found the reason:
Because I checked out the project on my Linux environment just recently, a newer version of prettier is installed (the latest -> 1.18.2 to be exact) but on my macOS environment it is only 1.14.0.
When I explicitly locked the version to 1.14.0 on Linux the different formatting disappeared.

@andywer maybe we should declare the latest version of prettier in `package.json` but lock it (remove the caret), so that this does not happen in the future?

PR #760, opened 2019-09-25
Use high transaction fee for multi-sig transactions

### LINKED ISSUE #531: Use high tx fee for multi-sig transactions
Whenever we create a new multi-sig transaction, don't just take the usual transaction fee, but twice that fee, minimum 10k stroops.

Background: Even if the currently required fee is low, it might not be anymore at the time that we actually submit the transaction to the horizon later. We can just pick a high fee, since the tx fee behavior has been changed with a CAP about half a year ago, so that the network always charges the lowest possible fee - the fee you set is basically just a maximum amount.

Little change on the side: Rename label "Fee" to "Maximum fee".

Follow-up of #501.

Marcel-Ebert:
- [x] Create `selectSmartMultisigTransactionFee()` function that uses either twice the amount of the usual fee or 10k stroops as fallback.
- [x] Rename the label from "Fee" to "Maximum Fee" in `<TransactionSummary>`

Closes #531.

### DIFF
--- a/src/components/Payment/CreatePaymentForm.tsx
+++ b/src/components/Payment/CreatePaymentForm.tsx
@@ -15,7 +15,7 @@ import { useIsMobile, RefStateObject } from "../../hooks/userinterface"
 import { renderFormFieldError } from "../../lib/errors"
 import { findMatchingBalanceLine, getAccountMinimumBalance, stringifyAsset } from "../../lib/stellar"
 import { isPublicKey, isStellarAddress } from "../../lib/stellar-address"
-import { createPaymentOperation, createTransaction, multisigMinimumFee } from "../../lib/transaction"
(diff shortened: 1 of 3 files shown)


src/components/Payment/CreatePaymentForm.tsx
```diff
@@ -176,7 +176,7 @@ function PaymentCreationForm(props: Props) {
     const tx = await createTransaction([payment], {
       accountData: props.accountData,
       memo: federationMemo.type !== "none" ? federationMemo : userMemo,
-      minTransactionFee: isMultisigTx ? multisigMinimumFee : 0,
```
andywer: Ohh, I didn't remember this. Looks like there is already something similar in place. The change of this PR would only be to select twice the smart fee if it's higher than the min. multi-sig fee, right?

Let's talk about it tomorrow.

PR #761, opened 2019-09-25
Export transitions from theme.ts

Marcel-Ebert:
- [x] Extract the two transitions and put them into a separate file
- [x] Export both transitions from `theme.ts`
- [x] Replace local transition definitions with imports from `theme.ts`

### DIFF
--- a/src/components/Dialog/Transitions.tsx
+++ b/src/components/Dialog/Transitions.tsx
@@ -1,0 +1,11 @@
+import React from "react"
+import { TransitionProps } from "@material-ui/core/transitions/transition"
+import Slide from "@material-ui/core/Slide"
+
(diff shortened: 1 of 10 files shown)


src/components/Dialog/Transitions.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import { TransitionProps } from "@material-ui/core/transitions/transition"
+import Slide from "@material-ui/core/Slide"
+
+export const SlideUpTransition = React.forwardRef((props: TransitionProps, ref) => (
```
andywer: Interesting idea - one more indirection, use case -> theme -> concrete transition -> abstract transition.

we need to be careful to not overdo the abstractioning, but this makes sense 👍

PR #781, opened 2019-10-11
Show update icon on all accounts view

### LINKED ISSUE #673: Show update icon on all-accounts view
Our users don't update the app regularly on desktop (on mobile the app stores auto-update them). We need to make it more obvious that they should update.

Depending on how #750 turns out (if auto updater works out of the box), we might not even need this story.

Follow-up of #560.

## To Do

- Add an update icon button to the all-accounts view, next to the settings icon button (cog wheel)
- Make it a bit flashy - pick a nice color
- Download update on click

Marcel-Ebert:
- [x] Implement auto-update IPC for electron
- [x] Add `updater` implementation for electron and stubs for cordova and web
- [x] Always set the `feedURL` of the `autoUpdater` in `checkForUpdates()` so that it is available when `startUpdatingWithoutInfo()` is called
- [x] Add an update button on top of the all-accounts page if an update is available

Closes #673.


andywer:
Please rebase/merge on top of latest `master`. Still need to test in detail, but looks fine so far!

PR #795, opened 2019-10-15
Link swipeable balance items to asset details

### LINKED ISSUE #788: Link swipeable balance items directly to asset details

andywer:
... and fix a bug when the asset details dialog would not be visible under certain conditions.

Closes #788.


andywer:
Not sure if we want to proceed this path right now. Closing for the time being.

PR #798, opened 2019-10-17
Fix no rounded corners of DialogActionsBox in Withdraw and Trading view

Marcel-Ebert:
- [x] To fix padding issues on iOS wrap the `<DialogActionsBox` with a `<Portal>` component in the withdraw view and with the `<ActionsContainer>` in the trading view
- [x] Slightly reduce the width of the `<DialogActionsBox>` on mobile
- [x] Align the "Add Asset" buttons in trading and withdraw view to the bottom right hand side

### DIFF
--- a/src/components/Dialog/Generic.tsx
+++ b/src/components/Dialog/Generic.tsx
@@ -38,9 +38,9 @@ const useActionButtonStyles = makeStyles(theme => ({
   mobileDialogActionsBox: {
     display: "flex",
     position: "fixed",
-    left: 0,
+    left: 8,
+    right: 8,
     bottom: 0,
-    width: "100%",
     backgroundColor: "#fcfcfc",
     justifyContent: "flex-end"
   },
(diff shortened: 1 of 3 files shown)


src/components/Dialog/Generic.tsx
```diff
@@ -38,9 +38,9 @@ const useActionButtonStyles = makeStyles(theme => ({
   mobileDialogActionsBox: {
     display: "flex",
     position: "fixed",
-    left: 0,
+    left: 8,
```
andywer: @Marcel-Ebert Can you remember why we made it 8px left and right, instead of 0 as before?

Marcel-Ebert: To make it smaller

andywer: Yeah, but why? 😅
Can't remember anymore...

Marcel-Ebert: Because you were unhappy with the space between the button(s) and the edge of the screen 🤐

PR #799, opened 2019-10-17
Make keyboard transition more natural on iOS

### LINKED ISSUE #790: Try KeyboardShrinksView in cordova config
See <https://cordova.apache.org/docs/en/3.0.0/guide/platforms/ios/config.html>. Might feel more natural than what we have right now? Looks a bit weird when the keyboard appears on iOS.

Marcel-Ebert:
- [x] Use `Keyboard.shrinkView(true)` to make the webview shrink the content instead of moving it to the top
- [x] Remove the old iPhone notch fix to prevent jumping of the screen content when hiding the keyboard
- [x] Add `flexShrink: 0` to style of upper toggle section in CreateAccount form to prevent squashed layout
- [x] Set `minHeight` to 300 in TradingForm to prevent wrapping of the layout when the keyboard is shown
- [x] Force a webkit repaint on resize events

The repaint is forced by changing the style of elements that use the `dialog-body` class. The `display` attribute is changed temporarily to trigger the repaint but the change gets reverted immediately (based on suggestions from [this post](https://stackoverflow.com/questions/3485365/how-can-i-force-webkit-to-redraw-repaint-to-propagate-style-changes)).

Closes #790.


Marcel-Ebert:
@andywer feel free to suggest any naming changes regarding the `rerender-listener.ts` file (the file itself and its contents).

Issue #796, opened 2019-10-16
Refactor IPC message passing / platform code


andywer:
The following changes concern all three platforms (electron, cordova, web). For web, let's implement the same interface, but just make it a local function call as everything happens in the same context.

## Concept

Let's have an abstract IPC messaging implementation that is uniform across all platforms. Each platform implements their message handlers.

Benefit: One common concept across all platforms and only one single react-app-side implementation for messaging across all platforms.

## To do

- Migrate all Cordova and Electron IPC messaging code to an `expose()`-like API (see <https://github.com/satoshipay/solar/blob/master/electron/lib/ipc.js>)
- Provide dull implementation for web build
- Refactor code in `src/platform/`: Only one implementation that feeds the IPC

## Notes

Electron:
- Preload script will now only blindly pass messages, not distinct functions per functionality

What to do if some calls are not supported by a platform:
- Take biometric auth (only in Cordova right now): The `isBiometricAuthAvailable()` returns `false` and if called, the `requestBiometricAuth()` will throw.

PR #803, opened 2019-10-18
Port electron platform code to typescript

### LINKED ISSUE #771: Port electron platform code to TypeScript
The code in `electron/` is the only code that is JS, not TS. Make it TS as a preparation for future code sharing between different platform code bases.

# To do

- `electron/lib/` becomes `electron/src/`
- git-ignore `electron/lib/`
- add `electron/tsconfig.json`
- move `electron/index.js` and `electron/development.js` to `electron/src/index.{prod,dev}.js`
- update npm scripts in main `package.json` (make sure that web build does not run electron tsc)

Marcel-Ebert:
- [x] Add `tsconfig.json` to `src/electron` folder
- [x] Add `electron/lib` directory to git-ignore
- [x] Move and rename `electron/development.js` and `electron/index.js` to `electron/src/index.{dev/prod}.ts`
- [x] Migrate the `electron/lib/*.js` files to typescript and place them in `electron/src/`
- [x] Bump versions of dependencies used by electron code (primarily because the newer versions offer type definitions)
- [x] Move `src/types/electron.d.ts` to `electron/types/types.d.ts` and add a type definition for the `electron-reload` module
- [x] Declare `index.prod.js` as main file in `package.json` (I did not see any usage of this file or rather its precursor so I assumed that it is not declared as `main` by accident?)
- [x] Use different path and options for `electron-reload` in `index.dev.ts`
- [x] Add `prebuild:electron:ts` and `prebuild:electron:ts:watch` scripts to `package.json` -> starts transpiling of typescript files (with or without watch mode)
- [x] Use the new 'prebuild' scripts in existing build and dev scripts

Closes #771.


electron/src/preload.ts
```diff
@@ -1,14 +1,15 @@
+// tslint:disable-next-line: no-var-requires
 const { ipcRenderer } = require("electron")
 const electronProcess = process
 
 let nextCommandID = 1
 
-function createCommand(commandType, eventType) {
-  return (...args) => {
+function createCommand<T>(commandType: string, eventType: string): () => Promise<T> {
```
andywer: We could also type the command function arguments here (as another generic type param). But since we are gonna get rid of this function in the course of #796, anyway, this is totally fine with me as is.

package.json
```diff
@@ -10,21 +10,23 @@
   },
   "homepage": "https://solarwallet.io/",
   "repository": "github:satoshipay/solar",
-  "main": "./electron/lib/app.js",
+  "main": "./electron/lib/index.prod.js",
```
Marcel-Ebert: I changed this to `index.prod.js` because I guess this is the right place to declare our production index file? I was not 100% sure about this, but I did not find any usages of our old `electron/index.js` so I thought this was by accident.

PR #818, opened 2019-10-31
Allow opening error messages in detail view

### LINKED ISSUE #813: Error toasts are cut off
Trim the error message cleanly with an ellipsis ("…") and allow opening the full error message in a dialog when you click on the error toast.

Also very valuable to be able to open the error, so it does not auto-hide after a few seconds.

Marcel-Ebert:
- [x] Trim notifications by showing ellipsis when message is too long
- [x] Add `onClick` function that opens a dialog with the message of the notification if `notification.onClick` is undefined

Closes #813.

### DIFF
--- a/src/components/NotificationContainer.tsx
+++ b/src/components/NotificationContainer.tsx
@@ -1,6 +1,8 @@
 import React from "react"
 import Snackbar, { SnackbarOrigin } from "@material-ui/core/Snackbar"
+import Dialog from "@material-ui/core/Dialog"
 import SnackbarContent from "@material-ui/core/SnackbarContent"
+import Typography from "@material-ui/core/Typography"
 import makeStyles from "@material-ui/core/styles/makeStyles"
 import CheckIcon from "@material-ui/icons/CheckCircle"
 import ErrorIcon from "@material-ui/icons/Error"
@@ -11,7 +13,11 @@ import green from "@material-ui/core/colors/green"
 import grey from "@material-ui/core/colors/grey"
 import { Notification, NotificationsContext, NotificationType } from "../context/notifications"
 import { useOnlineStatus } from "../hooks/util"
-import theme from "../theme"
+import theme, { FullscreenDialogTransition } from "../theme"
+import DialogBody from "./Dialog/DialogBody"
+import { DialogActionsBox, ActionButton } from "./Dialog/Generic"
+import { Box } from "./Layout/Box"
+import MainTitle from "./MainTitle"
 
 const icons: { [key in NotificationType]: React.ComponentType<any> } = {
   connection: OfflineBoltIcon,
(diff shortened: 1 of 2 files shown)


src/components/NotificationContainer.tsx
```diff
@@ -139,19 +149,49 @@ function NotificationsContainer() {
     lastShownNotification.current = latestNotificationItem
   }
 
-  return (
+  const showDialog = React.useCallback(() => {
```
andywer: Two things, also concerns `closeDialog()`:
- Rename the variable: When reading the code I am not so much interested in knowing how it will display things (dialog, a drawer or whatever), but rather *what* it will display. Better call it `showNotificationDetails` or so.
- Shorten two one line, that bit of code doesn't do enough to deserve so much visual space in the source file :) (you can also drop the `setDialogOpen` from the array if you want, since the React.useState() setters are always immutable and memoized)

src/components/NotificationContainer.tsx
```diff
@@ -139,19 +149,49 @@ function NotificationsContainer() {
...
+    setDialogOpen(true)
+  }, [setDialogOpen])
+
+  const closeDialog = React.useCallback(() => {
+    closeNotification()
+    setDialogOpen(false)
+  }, [setDialogOpen, closeNotification])
+
+  const dialog = (
```
andywer: Let's move that into a small memoized component of its own.

src/components/NotificationContainer.tsx
```diff
@@ -139,19 +149,49 @@ function NotificationsContainer() {
...
         open={open}
-        onClick={notification ? notification.onClick : undefined}
+        onClick={notification && notification.onClick ? notification.onClick : showDialog}
         onClose={closeNotification}
       />
       <OfflineNotification message="Offline" open={!isOnline} />
     </>
   )
+
+  return dialogOpen ? dialog : notificationJSX
```
andywer: As a general principle: We shouldn't render dialogs conditionally. This will definitely break the transition animation.

src/components/NotificationContainer.tsx
```diff
@@ -139,19 +149,49 @@ function NotificationsContainer() {
...
+    setDialogOpen(false)
+  }, [setDialogOpen, closeNotification])
+
+  const dialog = (
+    <Dialog fullScreen open={dialogOpen} onClose={closeDialog} TransitionComponent={FullscreenDialogTransition}>
+      <DialogBody
+        top={<MainTitle onBack={closeDialog} title={"Error"} />}
+        actions={
+          <DialogActionsBox>
+            <ActionButton autoFocus icon={<CheckIcon />} onClick={closeDialog} type="primary">
```
andywer: Let's drop the `<CheckIcon />`, too. I think it's better to not use any icon here, as there is no real action involved.

PR #823, opened 2019-11-06
[AFTER X-MAS] Refactor ipc message passing

### LINKED ISSUE #796: Refactor IPC message passing / platform code

ebma:
This PR introduces extensive changes regarding the way the IPC is handled.

- [x] Create a shared types directory and symlink it to `src` and `electron/src` to make the type definitions available in both
- [x] Introduce new type definitions for IPC messaging (in `shared/types/ipc.d.ts` and `electron/types/types.d.ts`)
- [x] Create a facade for platform specific IPC implementations in `ipc.ts`
- [x] Group access to the qr-reader component in `components.ts`
- [x] Refactor all electron, cordova and web implementations to use the new IPC messaging API

Closes #796.


Marcel-Ebert:
Building the storybook currently fails due to a reference to the 'web/key-store' in `.storybook/webpack.config.js`:
```
new webpack.NormalModuleReplacementPlugin(/\/platform\/key-store$/, resource => {
      resource.request = resource.request.replace("/platform/key-store", "/platform/web/key-store")
    }),
```
Removing this statement does not resolve the issue

andywer:
Rebased the feature branch and fixed the storybook build.

PR #824, opened 2019-11-06
Hide action buttons when keyboard is shown on android

Marcel-Ebert:
Hide the action buttons on android when the on-screen keyboard is shown (just like on iOS where this change was introduced recently).

- [x] Make the CSS media query selector for detection of open keyboard apply on iOS and android
- [x] Rename `iOSKeyboardHackSelector` to `MobileOnScreenKeyboardHackSelector` because it now applies to both mobile platforms


andywer:
Renamed the identifier to something that's a bit easier on the eyes 😉

PR #827, opened 2019-11-07
Don't show scrollbar on account title

### LINKED ISSUE #808: Scrollbar on account title is back on Windows
As reported by Alex:

image

Marcel-Ebert:
Fixes a bug where the scrollbar was shown for the account title div while in account settings.

The problem was that the buttons next to the account title when editing the name (Edit, Check and Cancel) had an overall height of 48px (24 normal + 12*2 through padding) but the height of the title div was set to 40px. 

I decided to increase the height to 48px instead of reducing the padding of the buttons because we would have to consider this padding change in the future when adding new buttons next to the account title component.

Closes #808.


andywer:
(Just happen to be awake right now and saw the notification:) Can you quickly try to drop the fixed height altogether and set a `overflow-y: visible` instead? Could potentially be a more future-proof solution:)

Marcel-Ebert:
@andywer do you mean `overflow-y: hidden`? 
It works but then the hover effect is slightly cropped:

2019-11-07-104942_881x100_scrot

andywer:
Why not `overflow-y: visible`? I would hope that it would just render the whole thing then, without scrollbars.

Marcel-Ebert:
Hmm I just noticed that the scrollbar is directly related to the `overflow-x: "hidden"` property set in 
https://github.com/satoshipay/solar/blob/813c6845b90ee5545e61d5a434e6e61182c7ac7b/src/components/MainTitle.tsx#L79
But the important part here is that `overflow-x` is set to hidden.
I tested this by explicitly setting `overflow-x: hidden` without `overflow-y` and the scrollbar still appears.

Any ideas why `overflow-x` would cause a vertical scrollbar to appear?

andywer:
I found this: https://stackoverflow.com/questions/6179831/setting-overflow-x-hidden-adds-a-vertical-scrollbar

But we don't have an `overflow-x: hidden` here, but a `overflow: hidden` which shouldn’t trigger that issue.

Marcel-Ebert:
I tried some things from the post you mentioned e.g. using a div as wrapper and a div for the content and splitting the overflow declarations between those two but I didn't have any luck. Or actually it worked by additionally setting `position: fixed` but this would break the layout on smaller screen width.

According to some comments on this issue, some combinations of `overflow-x` and `overflow-y` are not possible together and turned into something with `auto` automatically.

PR #839, opened 2019-11-11
Make Android status bar transparent 

ebma:
This PR introduces changes that will make the android status bar appear transparent. To be more precise, some flags are set so that the activity is shown in full screen mode while still showing system info in the status bar. The full screen mode requires the application to set additional padding on top so that the content does not overlap the info of the status bar. 

In addition to this, some layout changes to `<ReceivePaymentDialog>` and `<SettingsPage>` were made so that their styles align with those of the other components.
The upper-area of the settings-page was adjusted to match the styling of the header of the account-page.

### DIFF
--- a/src/components/Layout/Page.tsx
+++ b/src/components/Layout/Page.tsx
@@ -41,6 +41,7 @@ function PageInset(props: { children: React.ReactNode }) {
 }
 
 interface SectionProps extends BoxProps {
+  inheritBackground?: boolean
   backgroundColor?: React.CSSProperties["backgroundColor"]
   bottom?: boolean
   brandColored?: boolean
@@ -52,7 +53,11 @@ interface SectionProps extends BoxProps {
 }
 
 const Section = React.memo(function Section(props: SectionProps) {
-  const background = props.brandColored ? primaryBackground : props.backgroundColor || "#fcfcfc"
+  const background = props.brandColored
+    ? primaryBackground
+    : props.inheritBackground
+    ? undefined
+    : props.backgroundColor || "#fcfcfc"
   const isSmallScreen = useIsMobile()
 
   const className = [
--- a/src/components/Layout/Page.tsx
+++ b/src/components/Layout/Page.tsx
@@ -20,6 +20,9 @@ function TopOfTopSection(props: { background?: React.CSSProperties["background"]
   } else if (platform === "ios") {
     // Add some additional v-space for the iPhone X notch
     return <div className="iphone-notch-top-spacing" />
+  } else if (platform === "android") {
+    // Add some additional v-space for the android full screen
+    return <div className="android-top-spacing" />
   } else {
     return null
   }
@@ -54,7 +57,11 @@ const Section = React.memo(function Section(props: SectionProps) {
   const isSmallScreen = useIsMobile()
 
   const className = [
-    platform === "ios" && props.top ? "iphone-notch-top-spacing" : "",
+    platform === "ios" && props.top
+      ? "iphone-notch-top-spacing"
+      : platform === "android" && props.top
+      ? "android-top-spacing"
+      : "",
     platform === "ios" ? "iphone-notch-left-spacing" : "",
     platform === "ios" ? "iphone-notch-right-spacing" : "",
     platform === "ios" && props.bottom ? "iphone-notch-bottom-spacing" : ""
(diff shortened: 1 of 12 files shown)


src/components/Layout/Page.tsx
```diff
@@ -19,7 +21,10 @@ function TopOfTopSection(props: { background?: React.CSSProperties["background"]
     )
   } else if (platform === "ios") {
     // Add some additional v-space for the iPhone X notch
-    return <div className="iphone-notch-top-spacing" />
+    return isSmallScreen ? <div className="iphone-notch-top-spacing" /> : <></>
```
andywer: Shouldn't it show the top padding on larger iPad screens as well?

src/components/Layout/Page.tsx
```diff
@@ -35,30 +40,26 @@ function PageInset(props: { children: React.ReactNode }) {
   )
 }
 
-interface SectionProps {
-  children: React.ReactNode
+interface SectionProps extends BoxProps {
+  inheritBackground?: boolean
```
andywer: Does it really make sense to introduce the new `inheritBackground ` prop here? As far as I can tell you could achieve the same effect by passing `backgroundColor="inherit"`.

src/components/Layout/Page.tsx
```diff
@@ -35,30 +40,26 @@ function PageInset(props: { children: React.ReactNode }) {
...
 const Section = React.memo(function Section(props: SectionProps) {
-  const background = props.brandColored ? primaryBackground : props.backgroundColor || "#fcfcfc"
+  const background = props.brandColored
+    ? primaryBackground
+    : props.inheritBackground
+    ? undefined
+    : props.backgroundColor || "#fcfcfc"
   const isSmallScreen = useIsMobile()
 
-  const className = [
```
andywer: Why did you remove the `className` with the iOS notch classes? This looks like it's gonna break the layout on iOS now.

src/components/Payment/ReceivePaymentDialog.tsx
```diff
@@ -1,25 +1,22 @@
...
 }
 
 function ReceivePaymentDialog(props: Props) {
-  const isSmallScreen = useIsMobile()
-
   return (
-    <>
-      <Box width="100%" maxWidth={900} padding={isSmallScreen ? "24px" : " 24px 32px"} margin="0 auto 32px">
-        <MainTitle onBack={props.onClose} title="Receive Funds" />
+    <DialogBody top={<MainTitle onBack={props.onClose} title="Receive Funds" />}>
```
andywer: Good catch 👍

src/components/Layout/Page.tsx
Marcel-Ebert: No, the `isSmallScreen` is necessary because for example dialogs like the TransactionReview are not shown in fullScreen mode on larger screens so that without this check the 'non-fullscreen' dialogs would have too much padding on top.

src/components/Layout/Page.tsx
Marcel-Ebert: You're probably right that it makes more sense that way. 
I changed it so that `<DialogBody>` now has an optional `backgroundColor` instead of `inheritBackground` prop.

src/components/Layout/Page.tsx
Marcel-Ebert: I tested this on different virtual iOS devices (iPhone X 11.0.1, iPhone 11 Pro Max 13.2.2, iPhone 8 13.2.2, iPhone 7 11.0.1) and it looks just fine.
So I'd assume it is safe to remove it as I could not find any scenario where it seemed to change anything. 
But if you don't feel good or confident about removing it, we can keep it anyways.

andywer: Cool. Then let's try it.

Marcel-Ebert:
Good catch with that submission progress.
When I replaced the `<VerticalLayout>` with the `<Section>` in the `<DialogBody>` component I forgot to add the `display:"flex"` and `flexDirection:"column"` attributes (which a `<VerticalLayout>` usually has).

andywer:
I think we should look into this together on Wednesday.

ebma:
I rebased to the latest master again.
@andywer please have a look so we can merge it eventually.

PR #843, opened 2019-11-13
Show numeric keyboard when entering an amount

### LINKED ISSUE #826: Show numeric keyboard when entering an amount
Set `inputmode=decimal` on amount input fields. See <https://github.com/ionic-team/cordova-plugin-ionic-keyboard/issues/123>.

Concerns "send payment", "withdraw", "custom trustline" and potentially other dialogs.

Marcel-Ebert:
Add `pattern: "[0-9]*"` and `inputMode: "decimal"` to `inputProps` of TextFields that are supposed to handle numeric input.
This will cause the onscreen keyboards on Android and iOS to show numeric input keys.

Closes #826.

### DIFF
--- a/src/components/AccountAssets/CustomTrustline.tsx
+++ b/src/components/AccountAssets/CustomTrustline.tsx
@@ -51,6 +51,10 @@ function CustomTrustlineDialog(props: Props) {
           onChange={event => setIssuerPublicKey(event.target.value)}
         />
         <TextField
+          inputProps={{
+            pattern: "[0-9]*",
+            inputMode: "decimal"
+          }}
           fullWidth
           label="Limit (optional)"
           placeholder="Limit trust in this asset / maximum balance to hold"
--- a/src/components/Form/FormFields.tsx
+++ b/src/components/Form/FormFields.tsx
@@ -52,6 +52,10 @@ export const PriceInput = React.memo(function PriceInput(props: PriceInputProps)
   return (
     <TextField
       {...textfieldProps}
+      inputProps={{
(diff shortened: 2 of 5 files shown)


src/components/AccountAssets/CustomTrustline.tsx
```diff
@@ -51,6 +51,10 @@ function CustomTrustlineDialog(props: Props) {
           onChange={event => setIssuerPublicKey(event.target.value)}
         />
         <TextField
+          inputProps={{
+            pattern: "[0-9]*",
```
andywer: Using `pattern` seems dangerous. We need to allow entering decimal places, but depending on the user's locale we might run into trouble (in non-English locales a decimal point could also be a comma or a single quote, for instance).

Is there a good reason to use the pattern?

Marcel-Ebert: The pattern property is necessary to make it look like [this](https://user-images.githubusercontent.com/27284008/65964136-8ae68c80-e45c-11e9-9412-92ad1e3a4b0b.png).
Without it, the iOS keyboard won't show only the numbers but other characters as well, if I remember correctly.

andywer: True, it does show the decimal point. Just tested it!
Have you tried it on Android, too?

Marcel-Ebert: Yes I did. And I was pleasantly surprised because it has a similar effect on the Keyboard (showing a huge block with numbers) instead of showing a number row + normal letters.

PR #868, opened 2019-11-28
Display menus full-width on bottom (on mobile)

### LINKED ISSUE #842: [Mobile] Display menus full-width on bottom
Adapt the `theme.ts` to show `<Menu>` components full-width fixed to the bottom of the screen on mobile.

Make sure it still transitions nicely in and out.

~~Try setting the `native` prop for `<Select>` components via the theme when on mobile.~~ (Edit: Won't work as we'd need to render `<option>` instead of `<MenuItem>`, too)

Marcel-Ebert:
Change the styles in `theme.ts` so that menus like the `<AccountContextMenu>` and selectors in the trade and withdraw views are shown full-width at the bottom of the screen on mobile devices.

Closes #842.

### DIFF
--- a/src/components/Account/AccountContextMenu.tsx
+++ b/src/components/Account/AccountContextMenu.tsx
@@ -68,11 +68,17 @@ interface MenuProps {
 
 function AccountContextMenu(props: MenuProps) {
   const isSmallScreen = useIsMobile()
+
   return (
     <ContextMenu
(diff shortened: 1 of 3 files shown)


andywer:
Can merge this one. Can also wait a few days longer – Daniel will see if he has an idea how to polish the look & feel.

PR #870, opened 2019-11-28
Add spinner to TX review confirm button

### LINKED ISSUE #863: Missing spinner on tx review confirm button
Should show a spinner there, too, while the final preparations are done before submitting the tx to the network.

Usually it's pretty much instant, but there is some network communication going on (fetching fee stats?) that can take a moment when connectivity is bad.

ebma:
I used this small and simple fix which does not rely on a `txCreationPending` prop from the parent because the only time this will cause bad UX is when we go offline while in this view (as we cannot reach it if we were offline beforehand already). 

Closes #863.


Marcel-Ebert:
I memoized the onclick functions and icons now.
Furthermore I changed it so that the loading indicator is set to false once `onConfirm` has finished or an error was thrown.

PR #881, opened 2019-12-09
Highlight invalid input in the trading view

### LINKED ISSUE #875: Trading: Highlight invalid input
Highlight input fields with invalid data instead of disabling the "Place order" button.

Similar to #865 to make overall UI more consistent.

ebma:
- [x] Add `formik` dependency
- [x] Add `form` element
- [x] Move validation and submit code to initialisation of `useFormik` hook
- [x] Make the 'Place order' button always clickable
- [x] Add a `selectOnFocus` prop to the `<TradingPrice>` component and enable it in the `TradingForm` so that the content of the "Limit Price"-textfield is selected once focused which makes it easier to edit

Closes #875.

### DIFF
--- a/src/components/Trading/TradingForm.tsx
+++ b/src/components/Trading/TradingForm.tsx
@@ -367,6 +367,7 @@ function TradingForm(props: Props) {
                 priceDenotedIn={priceMode}
                 primaryAsset={primaryAsset}
                 secondaryAsset={secondaryAsset}
+                selectOnFocus
                 style={{ flexGrow: 1, maxWidth: 250, width: "55%" }}
               />
             </ExpansionPanelDetails>
--- a/src/components/Trading/TradingForm.tsx
+++ b/src/components/Trading/TradingForm.tsx
@@ -69,117 +69,24 @@ const useStyles = makeStyles({
   }
 })
 
-const prepareValidateFormFields = (props: Props, secondaryAmount: BigNumber, defaultPrice: string) => (
-  values: TradingFormValues
-) => {
-  const errors: FormikErrors<TradingFormValues> = {}
-
-  const { manualPrice, primaryAmountString, primaryAsset, secondaryAsset } = values
-
-  const primaryAmount =
-    primaryAmountString && isValidAmount(primaryAmountString) ? BigNumber(primaryAmountString) : BigNumber(0)
-
-  const primaryBalance = primaryAsset ? findMatchingBalance(props.accountData.balances, primaryAsset) : undefined
-  const secondaryBalance = secondaryAsset ? findMatchingBalance(props.accountData.balances, secondaryAsset) : undefined
-
-  if (!primaryAsset) {
-    errors.primaryAsset = "No asset selected."
-  }
-
-  if (!primaryAmountString) {
-    errors.primaryAmountString = "No amount specified!"
-  } else if (
-    primaryAmount.lt(0) ||
-    (primaryAmountString.length > 0 && primaryAmount.eq(0)) ||
-    (props.primaryAction === "sell" && primaryBalance && primaryAmount.gt(primaryBalance.balance)) ||
-    (props.primaryAction === "buy" && secondaryBalance && secondaryAmount.gt(secondaryBalance.balance))
-  ) {
-    errors.primaryAmountString = "Invalid amount specified!"
-  }
-
-  if ((manualPrice && (!isValidAmount(manualPrice) || BigNumber(manualPrice).eq(0))) || BigNumber(defaultPrice).eq(0)) {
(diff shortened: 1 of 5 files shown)


src/components/Trading/TradingForm.tsx
```diff
@@ -53,22 +58,88 @@ interface Props {
...
+
+  const formik = useFormik<TradingFormValues>({
+    initialValues: {
+      primaryAsset: props.initialPrimaryAsset,
+      primaryAmountString: "",
+      secondaryAsset: Asset.native(),
+      manualPrice: undefined,
+      priceMode: "secondary"
+    },
+    validate(values) {
```
andywer: We should not inline this validation function, but rather declare it before the component. Should be easy and reduces the component's size.

src/components/Trading/TradingForm.tsx
```diff
@@ -53,22 +58,88 @@ interface Props {
...
+      if (
+        (values.manualPrice && (!isValidAmount(values.manualPrice) || BigNumber(values.manualPrice).eq(0))) ||
+        BigNumber(defaultPrice).eq(0)
+      ) {
+        errors.manualPrice = "Invalid Price."
+      }
+
+      return errors
+    },
+    async onSubmit() {
```
andywer: Quick idea (and maybe not scope of this PR): How about

```ts
async onSubmit(values: TradingFormValues) {
  try {
    const tx = props.primaryAction === "buy"
      ? await createBuyTx(values, horizon, props.account, props.accountData)
      : await createSellTx(values, horizon, props.account, props.accountData)
    props.sendTransaction(tx)
  } catch (error) {
    trackError(error)
  }
}
```
?

andywer:
If we could only find a nice way to use `<Formik>` and its context… We would lose a bit of static typing, since we could not carry the generic types down through the context, BUT we could stop setting `onBlur`, `onChange` & `value` props manually and just wrap our form components in formik helpers that create these props based on the current form context.

Problem right now is that we want to be able to access the current form values in the `TradingForm` component, so we need the `useFormik()` hook and would then need to pour that current formik state into a formik context – we would need a `<FormikProvider>` component or similar that we can render in `TradingForm`.

(Probably not actionable right now, just sharing my thoughts)

andywer:
@ebma Please check out this PR. It needs a little more love ♥

src/components/Trading/TradingForm.tsx
andywer: 👆
This comment still needs to be addressed 😉

ebma: It's actually not that easy because the validation function uses other variables like the `secondaryAmount` which is calculated inside the `TradingForm` component and the calculation requires things like a `tradePair` which is accessed with a hook. So we cannot calculate the `secondaryAmount` in the validation function if it's outside the functional component since we can't get the tradepair. 

Do you have any ideas on how to circumvent this? Maybe I am just missing something.

andywer: Hmm, true. One solution could be currying:

```ts
const prepareValidateFormFields = (tradePair, secondaryAmount) => (values) => {
  // ...
}

// ...

const formik = useFormik<TradingFormValues>({
  // ...
  validate: prepareValidateFormFields(tradePair, secondaryAmount)
})
```

Another solution would be to move the validation method into a hook of its own, so it can use other hooks or we pass those secondary variables when we call the hook (`const validate = useTradingFormValidation(tradePair, secondaryAmount)`).

I think I like the hook-based approach better, but we can talk about it tomorrow.

ebma:
Rebased to latest master and added some changes.

I moved the validation function outside of the component but I'm not quite sure if how I did it makes it better or worse. 
I decided to use the currying method because we can't really use the hooks in this scenario because we need the `values` to initialise them and we only receive these in the actual function that is returned by the hook which on the other hand cannot use the hooks as this would violate the react hook rules.

What I'm trying to say is:
```
const useTradingFormValidation (props, tradePair, secondaryAmount) => 
// here we could use hooks
return (values) => {
// but only here we have the values to correctly initialise them
// which would cause an invariant hook violation
}
```

I still feel like I didn't figure out the best solution but I couldn't think of something better than how it is now.

andywer:
@ebma What hooks do we need for form validation that require `values` for initialization? And wouldn’t we then already have this issue now if two hooks were dependent on each other… 😉

ebma:
I moved the validation and calculation logic out of the `TradingForm` component.
Still no perfect solution but it works fine with `validateOnChange` set to `false` so let's not waste any more time on this and move on.

andywer:
We will use react-hook-form instead of Formik.

PR #883, opened 2019-12-09
Fix submission-progress not displayed properly

### LINKED ISSUE #877: TX error too big for submission progress dialog on desktop
When an error occurs after submitting a transaction, the displayed error content  is too big for the dialog so that the it is cut off and appears scrollable.

Screenshot 2019-12-05 at 16 12 08

Expected behaviour:
The dialog becomes big enough to fit the content of the error message without scrolling.

ebma:
The problem was that the `position` of the div containing the submission progress was set to 'absolute' and it seems like the dialog only adjusts to content with `position` set to 'relative'.

We needed that absolute position because we put the submission progress on top of the transaction details instead of hiding it. So in order to make the `position:"relative"` property work, we have to hide the transaction details when the submission progress is shown.

This was so annoying to fix/find out that I think this will haunt me in my dreams tonight.

Closes #877.

### DIFF
--- a/src/components/SubmissionProgress.tsx
+++ b/src/components/SubmissionProgress.tsx
@@ -50,7 +50,7 @@ function SubmissionProgress(props: SubmissionProgressProps) {
       promise={props.promise}
       pending={
         <Container>
-          <CircularProgress size={70} />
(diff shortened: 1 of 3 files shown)


andywer:
Good job! 🙌

PR #884, opened 2019-12-11
Highlight invalid input in the payment view

### LINKED ISSUE #865: Payment: Highlight invalid input
Highlight input fields with invalid data instead of disabling the "send" button.

ebma:
This branch is based on the changes made in #881. 

- [x] Replace existing validation handling in `PaymentForm` with implementation of `formik`
- [x] Merge `<PaymentForm>` and `<PaymentFormContainer>` 
- [x] Remove the `onCancel` prop from `<PaymentForm>` (as it is never used because the `<MainTitle>` component in the `<PaymentDialog>` is used to go back/cancel)

Closes #865.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12498,6 +12498,37 @@
       "integrity": "sha1-1hcBB+nv3E7TDJ3DkBbflCtctYs=",
       "dev": true
     },
+    "formik": {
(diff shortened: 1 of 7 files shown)


andywer:
We will use react-hook-form instead of Formik.

PR #888, opened 2019-12-18
Pass preselected asset when opening trading view from asset details

### LINKED ISSUE #857: Allow preselected asset when entering trade view
As reported by Alex:
"when I am in the asset detail view (e.g. BAT) and click TRADE, then SELL, I feel that it is a bug, that BAT is not prefilled in that trade screen."

## To Do

- Allow passing an asset to the trading view as part of the URL
- Preselect asset when linking from asset details view

ebma:
- [x] Change route for `tradeAsset`
- [x] Add asset-code of selected asset to url when clicking 'Trade' in asset-details view
- [x] Parse asset-code (if any) to asset in trading dialog and pass that asset to the trading form
- [x] Add `getLastArgumentFromURL()` function

The preselected asset will be preserved on back navigation so that if the user clicks on 'buy' first and decides that he rather wants to sell it, the asset will get preselected again. 

Closes #857.

### DIFF
--- a/src/components/Trading/TradingDialog.tsx
+++ b/src/components/Trading/TradingDialog.tsx
@@ -5,6 +5,7 @@ import Typography from "@material-ui/core/Typography"
 import { Account } from "../../context/accounts"
 import { useLiveAccountData } from "../../hooks/stellar-subscriptions"
 import { useDialogActions, useRouter } from "../../hooks/userinterface"
+import { getLastArgumentFromURL } from "../../lib/url"
 import { matchesRoute } from "../../lib/routes"
 import { parseAssetID, stringifyAsset } from "../../lib/stellar"
 import * as routes from "../../routes"
@@ -28,7 +29,7 @@ interface TradingDialogProps {
 
 function getAssetFromPath(pathname: string) {
   if (matchesRoute(pathname, routes.tradeAsset("*", undefined, "*"))) {
-    const lastArgument = pathname.replace(/^.*\/([^\/]+)/, "$1")
+    const lastArgument = getLastArgumentFromURL(pathname)
     if (lastArgument !== "buy" && lastArgument !== "sell") {
       return parseAssetID(lastArgument)
     }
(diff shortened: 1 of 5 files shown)


src/routes.ts
```diff
@@ -19,8 +19,8 @@ export const withdrawAsset = (accountID: string) => `/account/${accountID}/withd
 export const showTransaction = (accountID: string, transactionHash: string) =>
   `/account/${accountID}/tx/${transactionHash}`
 
-export const tradeAsset = (accountID: string, method?: "buy" | "sell") => {
-  return method ? `/account/${accountID}/trade/${method}` : `/account/${accountID}/trade`
+export const tradeAsset = (accountID: string, method?: "buy" | "sell", preselectedAsset?: string) => {
+  return `/account/${accountID}/trade` + (method ? `/${method}` : "") + (preselectedAsset ? `/${preselectedAsset}` : "")
```
andywer: How about this?
```suggestion
  return [
    `/account/${accountID}/trade`,
    method || "",
    preselectedAsset || ""
  ].filter(fragment => !!fragment).join("/")
```

src/components/Trading/TradingDialog.tsx
```diff
@@ -24,10 +26,26 @@ interface TradingDialogProps {
   sendTransaction: (transaction: Transaction) => void
 }
 
+function getAssetFromPath(pathname: string) {
+  if (matchesRoute(pathname, routes.tradeAsset("*", undefined, "*"))) {
```
andywer: [Super minor] Only change this if you need to touch things here once more anyway:

That kind of code is usually better readable / comes with less cognitive overhead if you turn the `if X then some block of Y` into `if not X exit early; some block of Y` 🙂

andywer:
Just tested it – works nicely 🚀

src/routes.ts
ebma: Definitely looks more readable than the string concatenation I used. 👍

src/components/Trading/TradingDialog.tsx
ebma: I decided not to change it because in this case we'd need two `return undefined` statements since the function returns `undefined` when either the first or the second `if`-statement is not true.

But you're probably right that it's more intuitive to exit early and I will try to keep it in mind and use it in future scenarios :)

ebma:
I rebased this branch and resolved two conflicts.

src/components/Trading/TradingDialog.tsx
andywer: Fair enough 😉

PR #889, opened 2019-12-18
[WIP] Use BitPay QR code scanner

### LINKED ISSUE #887: Try BitPay QR code scanner
Our current QR code scanner for Cordova isn't too pretty and feels laggy, too.

Let’s try <https://github.com/bitpay/cordova-plugin-qrscanner> instead.

ebma:
I investigated the usage of this plugin and hacked some stuff together which is still WIP.

Quick summary: 
The plugin works different to the one we are currently using. It does not start a separate activity/view but the camera kind of appears behind the web content. That's why we need to hide our current content in order to make the camera show. 
The camera view of the plugin does not provide any UI elements which might be a good thing since we can now add our own custom UI (which is not possible with our current plugin because it starts the new activity). I already added a custom back button (just a small, clickable div) but we should think about how we want to handle the back navigation. 

Another advantage is that we can basically already open the camera in the background before the user actually clicked the 'scan QR-code'-Button and just hide the other content when the button is clicked so that the camera appears to open super fast (although it actually was open the whole time).

I couldn't test if it works on iOS as I don't have access to a physical iPhone and the emulator does not support camera usage but I found an [issue ticket](https://github.com/bitpay/cordova-plugin-qrscanner/issues/253#issuecomment-488720284) with information that might come in handy when we encounter problems on iOS.

TODO: 
- [x] Add `cordova-plugin-qrscanner`
- [x] Use the new qr-scanner plugin in cordova
- [ ] Test on iOS
- [ ] Add/Change UI elements 
- [ ] Fix android back navigation (when physical back button is clicked it navigates back instead of closing the camera)

Closes #887

### DIFF
--- a/cordova/package-lock.json
+++ b/cordova/package-lock.json
@@ -827,6 +827,15 @@
       "resolved": "https://registry.npmjs.org/cordova-plugin-network-information/-/cordova-plugin-network-information-2.0.2.tgz",
       "integrity": "sha512-NwO3qDBNL/vJxUxBTPNOA1HvkDf9eTeGH8JSZiwy1jq2W2mJKQEDBwqWkaEQS19Yd/MQTiw0cykxg5D7u4J6cQ=="
     },
+    "cordova-plugin-qrscanner": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/cordova-plugin-qrscanner/-/cordova-plugin-qrscanner-3.0.1.tgz",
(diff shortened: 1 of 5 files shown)


ebma:
@andywer could you give it a quick try on iOS and decide whether the performance improvements justify replacing our current plugin with this one?

andywer:
No priority, unfortunately. Closing for now.

PR #901, opened 2020-01-06
Fix wrong amount shown in order description

### LINKED ISSUE #891: Wrong quantity displayed on Recent transactions
[I am totally new on GitHub, I hope this report is correctly placed here and with this format]
buy and sell rates
buy and sell rates - detail

It seems that the quantity of XLM going to be sold in order to buy EURT is wrongly calculated. Instead, it is correctly displayed on Open orders.

This happened in Testnet, I don't know if it happens in public too.

ebma:
Fix wrong amount shown in the description of orders in the recent transaction list.

@andywer I assume that changing the calculation from `div` to `mul` is correct because we do the same for the `TransactionReview` (?)
https://github.com/satoshipay/solar/blob/4085be3b0b225ab375799f11c494c0aa73174775/src/components/TransactionReview/Operations.tsx#L213-L216

Closes #891.

### DIFF
--- a/src/components/Account/TransactionList.tsx
+++ b/src/components/Account/TransactionList.tsx
@@ -77,7 +77,7 @@ function OfferDescription(props: {
     <>
       {prefix}
       {props.type === "manageBuyOffer"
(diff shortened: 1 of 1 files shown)


andywer:
Looks good to me!

PR #910, opened 2020-01-09
Add support for kraken-style QR codes for payments

### LINKED ISSUE #897: Support kraken-style QR codes for payments
kraken.com deposit address QR codes encode payment data in this format:

```
<account-id>?dt=<memo>
```

Example: `GABC…XYZ?dt=123456789`

We need to parse the query params, use only the public key portion as the destination and use the `dt` parameter value as the memo value.

ebma:
The result of the QR code scan is split into parts at the `?` character.
 If the result contains additional params after the `?`, we specifically look for a `dt` parameter and set the `memoValue` according to it.

Closes #897.

### DIFF
--- a/src/components/Payment/PaymentForm.tsx
+++ b/src/components/Payment/PaymentForm.tsx
@@ -158,10 +158,29 @@ const PaymentForm = React.memo(function PaymentForm(props: PaymentFormProps) {
     props.onSubmit(formValues, spendableBalance, matchingWellknownAccount)
   }
 
+  const handleQRScan = React.useCallback((scanResult: string) => {
+    const array = scanResult.split("?")
+
+    const destination = array[0]
+    setFormValue("destination", destination)
+
+    if (array.length === 1) {
+      return
+    }
+
+    const searchParams = new URLSearchParams(array[1])
+    const memoValue = searchParams.get("dt")
+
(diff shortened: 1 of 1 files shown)


src/components/Payment/PaymentForm.tsx
```diff
@@ -158,10 +158,29 @@ const PaymentForm = React.memo(function PaymentForm(props: PaymentFormProps) {
     props.onSubmit(formValues, spendableBalance, matchingWellknownAccount)
   }
 
+  const handleQRScan = React.useCallback((scanResult: string) => {
```
andywer: We can write this function in a more elegant way (sorry, but the one-click suggestions in comments don't seem to work if you want to change more than one line…)

```ts
  const handleQRScan = React.useCallback((scanResult: string) => {
    const [destination, query] = scanResult.split("?")
    setFormValue("destination", destination)

    if (!query) {
      return
    }

    const searchParams = new URLSearchParams(query)
    const memoValue = searchParams.get("dt")

    // ...
}, [])
```

andywer: Looks fine otherwise – still need to test it.

ebma: I added your suggested change.

PR #911, opened 2020-01-09
[WIP] Add 'useTranslation'-hook and replace hard-coded strings

### LINKED ISSUE #876: Prepare for i18n: Replace hard-coded strings
Replace all hard-coded strings in the UI by a call to a translation function `t()` that looks up the localized string from a JSON object (hard-coded for now) that maps translation keys to translated messages.

Create a React hook to obtain the translation function `t()`, stick to the API of <https://react.i18next.com/latest/usetranslation-hook>, so we can easily switch to i18next or another i18n solution later.

## To be defined

- Do we already need i18next for parameter substitution in strings?

ebma:
- [x] Add `useTranslation()` hook (similar to the one from [react-i18next](https://react.i18next.com/latest/usetranslation-hook))
- [x] Start adding translations for the `en`-locale
- [ ] Add remaining translations for `en`-locale for all hard-coded strings
- [ ] Escape HTML-tags from strings if present

I decided to have some basic 'namespace' support as they do in `i18next` so that the translations-object becomes more readable. I think this is a good way to cluster it into logical units so that we can keep a better overview.

@andywer please let me know what you think and if you have some suggestions for improvements before I start adding the remaining translations.

Closes #876.

### DIFF
--- a/src/components/Account/AccountContextMenu.tsx
+++ b/src/components/Account/AccountContextMenu.tsx
@@ -15,6 +15,7 @@ import { Account } from "../../context/accounts"
 import { SettingsContextType } from "../../context/settings"
 import { useLiveAccountData } from "../../hooks/stellar-subscriptions"
 import { useIsMobile } from "../../hooks/userinterface"
+import { useTranslation } from "../../hooks/i18n"
 import ContextMenu, { AnchorRenderProps } from "../ContextMenu"
 
 const useContextMenuItemStyles = makeStyles({
@@ -72,6 +73,8 @@ interface MenuProps {
 function AccountContextMenu(props: MenuProps) {
   const accountData = useLiveAccountData(props.account.publicKey, props.account.testnet)
   const isSmallScreen = useIsMobile()
+  const { t } = useTranslation()
+
   const activated = accountData.balances.length > 0
 
   return (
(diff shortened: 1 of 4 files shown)


src/components/Account/AccountContextMenu.tsx
```diff
@@ -72,6 +73,8 @@ interface MenuProps {
 function AccountContextMenu(props: MenuProps) {
   const accountData = useLiveAccountData(props.account.publicKey, props.account.testnet)
   const isSmallScreen = useIsMobile()
+  const { t } = useTranslation()
```
andywer: Since we are only gonna use messages on the `account-context-menu` namespace, let's [scope `t` to this namespace](https://react.i18next.com/latest/usetranslation-hook#loading-namespaces):

```ts
  const { t } = useTranslation("account-context-menu")
```

src/hooks/i18n.ts
```diff
@@ -0,0 +1,93 @@
+import React from "react"
```
andywer: It's a nice little exercise to write something like this ourselves, but I think we can just use the actual `react-i18next` and just shove the translations in it. This way we immediately have support for parameter substitution, components in translations, etc. 😉

src/pages/all-accounts.tsx
```diff
@@ -43,9 +44,15 @@ function AllAccountsPage() {
   const styles = useStyles()
   const isSmallScreen = useIsMobile()
   const isWidthMax450 = useMediaQuery("(max-width:450px)")
+  const { t } = useTranslation()
 
-  const switchToMainnetLabel = isSmallScreen ? "Mainnet" : "Switch To Mainnet"
-  const switchToTestnetLabel = isSmallScreen ? "Testnet" : "Switch To Testnet"
+  const switchToMainnetLabel = isSmallScreen
+    ? t("all-accounts-page.switch-to-mainnet-label-short")
```
andywer: These translation keys look somewhat bulky…

How about `page.all-accounts.actions.switch-to-mainnet.label.{long,short}` and scoping the `useTranslation()` to `pages.all-accounts`?

src/pages/settings.tsx
```diff
@@ -32,66 +34,65 @@ function Settings() {
         checked={settings.biometricLock && settings.biometricLockUsable}
         onChange={settings.toggleBiometricLock}
         style={bioLockAvailable ? {} : { display: "none" }}
-        title={process.env.PLATFORM === "ios" ? "Face ID / Touch ID" : "Fingerprint Lock"}
+        title={
+          process.env.PLATFORM === "ios"
+            ? t("settings-page.biometric-lock-title-ios")
```
andywer: `page.settings.biometric.title.ios`+ `page.settings.biometric.title.default`?

src/pages/settings.tsx
```diff
@@ -116,7 +119,7 @@ function SettingsPage() {
             <MainTitle
               onBack={() => router.history.push(routes.allAccounts())}
               style={{ margin: "-12px 0 -10px", minHeight: 56 }}
-              title="Settings"
+              title={t("settings-page.main-title")}
```
andywer: `page.settings.title`?

src/components/Account/AccountContextMenu.tsx
andywer: (`page.account.context-menu` or similar if we go by the naming scheme I propose below)

src/pages/all-accounts.tsx
ebma: I like your idea of scoping with the `useTranslation` functions but I found some constraints while testing. 

Namespaces don't support nesting with the `.` syntax like the `keys` do.  This means that we could only scope it to `page` and not to `page.all-accounts` which is still an improvement but not as much as it could be.

For example we could use 
```
const { t } = useTranslation("page")
t("all-accounts.actions.switch-to-mainnet.label.short")
```
instead of 
```
const { t } = useTranslation()
t("page:all-accounts.actions.switch-to-mainnet.label.short")
```
with a `translation.json`:
```
{
   "page": {
    "all-accounts": {
      "actions": {
        "switch-to-mainnet": {
          "label": {
            "long": "Switch To Mainnet",
            "short": "Mainnet"
          }
        },
        "switch-to-testnet": {
          "label": {
            "long": "Switch To Testnet",
            "short": "Testnet"
          }
        }
      }
    }
  }
}
```


There is an example in [this](https://www.i18next.com/principles/namespaces#semantic-reasons) section from the docs. Maybe we should consider doing something similar to this?
I'm fine with both but I just wanted to point out some alternatives before we start adding the translations as it probably makes more sense thought about this beforehand.

ebma:
This PR will be split into smaller PRs where one adds the logic+configuration for `i18n` (see #915) and multiple others that replace the hard-coded strings over time.

andywer:
Let's close this PR and open PRs de-hardcoding strings in views. We just need to agree on a scheme for the translation keys. Let's move that to the story (#876), though.

PR #915, opened 2020-01-15
Add and configure 'i18next' 

ebma:
This PR adds the foundation for i18n support (described in #876).

- [x] Add `i18next`, `react-i18next` and `i18next-browser-languagedetector`
- [x] Add empty resource file for english translations
- [x] Configure `i18n` on app start


i18n/locales/en/translation.json
```diff
@@ -0,0 +1 @@
+{}
```
andywer: Why not just call it `i18n/locales/en.json`? 😉

ebma: I used the same structure as in this [example](https://github.com/i18next/react-i18next/blob/master/example/react/public/locales/en/translation.json) from `react-i18next`. 

I guess the only benefit is that we could split our translations into multiple files for different namespaces,  e.g. have a file called `translation.json` and another one called `common.json` where the naming represents the namespace so that we would have a 'cleaner' structure. 

But I don't know if we want to do this anyways so I don't mind if we chose a different path here.

andywer: Yeah, then let's put everything in a `en.json` for now – we can still split it up in multiple files later if it grows too large (and we shouldn't yet try to guess how we are gonna split it up later).

PR #916, opened 2020-01-16
Restyle app settings

ebma:
This PR changes the style of the app settings so that they look similar to the account settings.

- [x] Add an `AppSettings` component which is similar to the existing `AccountSettings` component
- [x] Add toggle sections to enable/disable every settings
- [x] Adjust the style of the header of the `settings`-page so that it matches alignment of the `account`-page header

Previews:
Screenshot 2020-01-23 at 17 42 18

Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-23 at 17 42 03


andywer:
Made some very small adjustments. Good job!

PR #918, opened 2020-01-20
Hide window menu on Linux (+ Windows?)

andywer:
`window.removeMenu()` turns out to not have done the job reliably.


andywer:
@andrenarchy confirmed that this change fixes it 🎉

PR #937, opened 2020-02-05
Fix 'wrong-password-error' not handled properly

ebma:
Fixes a bug where the `WrongPasswordError` was re-thrown in the `TransactionSender` so that an uncaught promise rejection occurred.


andywer:
Good job! 👍

PR #945, opened 2020-02-12
Translate errors

ebma:
This PR replaces hard-coded strings in validation errors and adds a new convention for 'general'/custom errors. When throwing a new Error which is not related to validation (e.g. in a form) one should throw a `CustomError` instead, where the `name` property maps to the translation key `error.{name-in-kebab-case}`.

- [x] Replace hard-coded strings of errors
- [x] Add `lodash.pick`
- [x] Update to version `1.1.0` of `threads.js` to get access to serialization
- [x] Add `CustomError` which wraps an existing error but can hold additional parameters that can be used for translation
- [x] Implement serialization for `CustomError`
- [x] Change error forwarding in cordova ipc (enable passing `extras` and `__extraProps`)


electron/src/ipc/_ipc.ts
```diff
@@ -1,4 +1,5 @@
 import { ipcMain } from "electron"
+import pick = require("lodash.pick")
```
andywer: Is there a particular reason for this `import pick = require()` syntax? Should work without.

ebma: No there is no reason, I guess it was added by my VSCode while auto-importing.

andywer:
We should merge this after v0.24.0 is out – QA on v0.24.0 is already quite far.

andywer:
Rebased the branch.

PR #952, opened 2020-02-17
Add eslint and configure it for use with react

### LINKED ISSUE #941: Add eslint
Run eslint alongside with tslint.

Main motivation: Use the rules that check for missing parameters in hooks' dependency arrays (like the 2nd argument to `useEffect()`).

ebma:
- [x] Add `eslint` + eslint-plugin dependencies
- [x] Add `.eslintrc.json` config which makes eslint only look for react related issues
- [x] Add call to eslint in `lint:fix` command in `package.json`
- [x] Fix errors depicted by `eslint`

Closes #941. 

**Note:** There are a total of 26 errors outlined and issues (related to hook-dependencies) across 53 files automatically fixed by eslint, so it would probably make sense to ~create a follow-up task to this one where these issues are fixed~ fix them right away.


package.json
```diff
@@ -13,7 +13,7 @@
   "main": "./electron/lib/index.prod.js",
   "scripts": {
     "precommit": "lint-staged",
-    "lint:fix": "tslint --fix '{src,stories}/**/*.{ts,tsx}'",
+    "lint:fix": "tslint --fix '{src,stories}/**/*.{ts,tsx}' && eslint '{src,stories,electron/src}/**/*.{ts,tsx}' --quiet --fix --no-eslintrc -c ./.eslintrc.json",
```
andywer: I never used `lint:fix`, I think. Was just part of some boilerplate of ours if I'm not mistaken.

**What would be way more important is to actually lint the code 😉**

Please adapt the `test` script.

ebma:
I fixed all errors by applying the `eslint --fix` first and fixing the remaining errors manually afterwards.

One arguable thing is that react rules of eslint want to have the whole `props` as dependency in the dependency array when a function contained in those props is called directly without creating another reference to it. This is intentional and not necessarily a bug (it's being discussed [here](https://github.com/facebook/react/issues/16265)).

I don't have any hard feelings about this and I don't mind putting the whole props object in the dependency array but I don't know if you do @andywer?

andywer:
Interesting. In those cases (see the linked discussion of your prev comment) we should probably destructure the props beforehand and pass those destructured props to the deps array. Might be the best solution here.

Feel free to object, though :)

ebma:
I destructured only the functions of the props that are used within `useEffect` or `useMemo` beforehand so that we solve the `exhaustive-deps` issue but still have a clear indication that the properties used from within the `useEffect` or `useMemo` belong to the props object.

PR #971, opened 2020-03-03
Split English translation file into multiple files

### LINKED ISSUE #948: Should we split the translation file?
Right now we have one big JSON file containing all the English messages only. It is already > 600 lines long. I think I argued before to "chop down the messages file once it becomes too big".

We might be approaching that point sooner than I thought… So here is my idea:

How about having one central messages file containing all the generic messages and then have translation files in `src/components/*/<lang>.json`?

Pro:
- Organizational structure is better scalable

Con:
- ?

Left to figure out:
- A way how to glue all those JSON fragments back to together into one global translations object without having to replicate all the imports from the component directories for each language

@ebma

andywer:
Closes #948. Did not put the translations in `src/components`, but kept them at `i18n/locales/en/*.json`.


andywer:
Should maybe wait to merge this PR until v0.24.0 is released.

PR #977, opened 2020-03-04
Create zh-CN.json

fm612:
chinese language

### DIFF
--- a/i18n/locales/zh-CN.json
+++ b/i18n/locales/zh-CN.json
@@ -1,0 +1,669 @@
+{
+  "accounts-context-menu": {
+    "account-settings": {
+      "label": "用户"
+    },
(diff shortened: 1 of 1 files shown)


fm612:
canot merging

andywer:
Hey @fm612!

We took the translations you send via email and one of our Chinese-speaking employees is reviewing it already. Let's open a new PR once he is done.

We also need to match the translations against the updated translation file structure (see #971).

Thanks for your awesome contribution! I can also send the updated translations back to you once they are updated and match #971, so you could open a new PR if you would like your GitHub account to appear as the creator of these translations :)

andywer:
tldr; Only closing the PR temporarily. We will re-open it once they are reviewed and we have updated them to match the new file structure.

PR #982, opened 2020-03-05
Patch 2

fm612:
chinese language


andywer:
Hey @fm612, please see my comment here: https://github.com/satoshipay/solar/pull/977#issuecomment-594430155

We are super happy about your contribution, but you need to be a bit patient while we are preparing everything. Opening so many PRs with single file changes is problematic – we need one combined PR. Please wait until your translations have been reviewed. Should be done by early next week.

We will ship a new release in the next days, but we cannot release the translations yet anyway, since most of, but not all of the user interface has been made translatable yet. We need a little bit more time 😉

andywer:
@fm612 If you really want, you can already open a PR that adds the separate Chinese translation files to #971, but then as one PR adding all the Chinese translations.

We can then apply changes afterwards.

PR #988, opened 2020-03-05
Create zh-CN.ts

fm612:


andywer:
See https://github.com/satoshipay/solar/pull/982#issuecomment-595119994

PR #980, opened 2020-03-04
Fix windows auto-updater

ebma:
Supersedes #943.

Fixes the windows auto-update mechanism by using `squirrel` as target package type for Windows.

- [x] Update `electron-builder` and `electron` dependency versions
- [x] ~Remove~ Update local electron type definitions
- [x] Add `electron-builder-squirrel-windows` as dependency
- [x] Configure `squirrel` as target in `electron-build.yml`
- [x] Cache the update availability status in the settings context
- [x] Test if it still works on macOS
- [x] (Optional) Add different installer gif (see [loadingGif](https://www.electron.build/configuration/squirrel-windows))

Important notes (for windows users): 
Since this changes the package type and the installer that comes with it, existing users will have two applications installed alongside each other. The old application will not be overwritten as the new installation directory is different and it will not be uninstalled automatically. This means that the user should manually uninstall the old application. 
I tested it on my windows device and the start menu and desktop shortcut are created successfully and are pointing to the new application BUT the old start menu shortcut file is in a directory above the new one thus when the user searches for 'Solar Wallet' via his start menu he will open the old application. That's why we should encourage the uninstallation of the old application. Doing so should not delete existing data (I tested it on two different devices running Win10). 

Closes #938.

PR #993, opened 2020-03-06
[WIP] Replace electron-builder with electron-forge

andywer:
Electron-forge seems to be the better choice: Cleaner configuration, modular, much better maintained, actively maintained squirrel.windows backend.

- [x] Basic electron-forge setup
- [ ] Code signing
- [ ] Update CI config
- [ ] Thorough testing / QA

Based on #980.


andywer:
Neither Windows nor Mac build currently works on a macOS machine if you want to sign them.

#### Mac build

Waits for an eternity for the notarization to complete, even after it seems to have completed already (received notarization success email from Apple minutes ago).

#### Windows build

Fails on macOS despite wine and mono with:

```
Failed to sign, command invoked was: '/Users/work/workspace/solar/node_modules/electron-winstaller/vendor/signtool.exe sign /a /f "/Users/work/Certificates/SatoshiPayLtd.p12" /p "<supersecretpassword>" /Users/work/.local/share/SquirrelTemp/tempa/lib/net45/d3dcompiler_47.dll'
```

There are multiple error messages like this for multiple files it tries to sign. Tried removing the `--arch=ia32` argument, too – no difference.

Here is a good issue with a fix that doesn't seem to be directly applicable to this electron-forge update, though: https://github.com/electron/electron/issues/7476

andywer:
Won't fix for the time being.

PR #995, opened 2020-03-09
Disable app.setAsDefaultProtocolClient() for now

andywer:
Until we actually roll-out SEP-7 support.

### DIFF
--- a/electron/src/bootstrap.ts
+++ b/electron/src/bootstrap.ts
@@ -1,5 +1,7 @@
 import { app } from "electron"
+
 // Needs to match the value in electron-build.yml
 app.setAppUserModelId("io.solarwallet.app")
 
-app.setAsDefaultProtocolClient("web+stellar")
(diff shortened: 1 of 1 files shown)


tourBermuda:
First I just wanted to say thanks for one of my favorite wallets. Can you give me an idea when sep-007 will be available on Solar wallet?

andywer:
Hey @tourBermuda, that's always very nice to hear!

Hard to tell right now as we have a plenty-full pipeline of things we need to work on and the focus points are currently pretty volatile as we have some external dependencies there. I think I will know more by the end of the week. Feel free to ping me again then 😉

A rough ballpark figure would be something between in four weeks to eight weeks maybe.

tourBermuda:
@andywer thanks. That's actually sooner than I thought. I'll touch base with you in about a month or two.  Please add the sep7 pay and tx transactions. We'll chat later.

PR #1001, opened 2020-03-12
Use react-hook-form to highlight invalid input in trading view

### LINKED ISSUE #875: Trading: Highlight invalid input

ebma:
Supersedes #881.

- [x] Add `react-hook-form` dependency
- [x] Add inputError and name prop to `AssetSelector`
- [x] Make TradingPrice accept a `defaultPrice` prop 
- [x] Move calculation into separate hook outside of `TradingForm` component 
- [x] Use `react-hook-form` for form validation

The `defaultPrice` prop was necessary because `react-hook-form` is handling value changes so that 
`manualPrice={manualPrice.value !== undefined ? manualPrice.value : defaultPrice}` won't work anymore. The `defaultPrice` will only be passed to the TradingPrice component if it has not been 'touched' yet which means that the user did change the content of that field.

Closes #875

### DIFF
--- a/src/components/Trading/TradingForm.tsx
+++ b/src/components/Trading/TradingForm.tsx
@@ -313,7 +313,6 @@ function TradingForm(props: Props) {
                 }
                 minWidth={75}
                 showXLM
-                onChange={() => undefined}
                 style={{ flexGrow: 1, marginRight: 24, maxWidth: 150, width: "25%" }}
                 testnet={props.account.testnet}
                 value={primaryAsset}
@@ -394,7 +393,6 @@ function TradingForm(props: Props) {
                 }
                 minWidth={75}
                 showXLM
-                onChange={() => undefined}
                 style={{ flexGrow: 1, marginRight: 24, maxWidth: 150, width: "25%" }}
                 testnet={props.account.testnet}
                 value={secondaryAsset}
--- a/src/components/Trading/TradingForm.tsx
+++ b/src/components/Trading/TradingForm.tsx
@@ -438,19 +438,22 @@ function TradingForm(props: Props) {
           </ExpansionPanelSummary>
           <ExpansionPanelDetails className={classes.expansionPanelDetails}>
             <Controller
-              as={TradingPrice}
+              as={
+                <TradingPrice
+                  defaultPrice={!form.formState.touched.manualPrice ? defaultPrice : undefined}
+                  inputError={form.errors.manualPrice && form.errors.manualPrice.message}
+                  onSetPriceDenotedIn={setPriceMode}
+                  price={effectivePrice}
+                  priceDenotedIn={priceMode}
+                  primaryAsset={primaryAsset}
+                  secondaryAsset={secondaryAsset}
+                  selectOnFocus
+                  style={{ flexGrow: 1, maxWidth: 250, width: "55%" }}
+                />
+              }
               control={form.control}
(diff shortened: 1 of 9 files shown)


src/components/Trading/TradingForm.tsx
```diff
@@ -247,25 +295,36 @@ function TradingForm(props: Props) {
         width="100%"
       >
         <HorizontalLayout margin="8px 0">
-          <AssetSelector
```
andywer: [Minor] Couldn't we add an `inputRef` prop to `AssetSelector` and use the `register()` call that we use for other inputs? It's quite hard to tell what props belong to the `Controller` and which ones belong to the `AssetSelector` now.

src/components/Trading/TradingForm.tsx
```diff
@@ -160,22 +131,99 @@ function TradingForm(props: Props) {
...
+function TradingForm(props: Props) {
+  const classes = useStyles()
+  const isSmallScreen = useIsMobile()
+  const isSmallHeightScreen = useMediaQuery("(max-height: 500px)")
+  const isSmallScreenXY = isSmallScreen || isSmallHeightScreen
+
+  const [expanded, setExpanded] = React.useState(false)
+  const [priceMode, setPriceMode] = React.useState<"primary" | "secondary">("secondary")
+
+  const { control, errors, formState, getValues, handleSubmit, register, setValue, triggerValidation, watch } = useForm<
```
andywer: [Minor] Just a quick thought: Since the returned object is quite complex and we need a lot of its props it might be worth trying a slightly different style of using it.

Let's try to not destructure it: `const form = useForm<TradingFormValues>(/*...*/)` and then later access things as `form.errors`, etc.

src/components/Trading/TradingForm.tsx
ebma: I now moved the props from the `<Controller>` back to the actual component so that it's clear where they belong. 
I think we should still make use of the `<Controller>` as it fits this use case pretty well.

Besides that I tried your suggested approach with passing an `inputRef` but this caused issues with the value being `undefined` for some renders which caused our `tradePair` hook to fail. Might be my bad but anyways, I think that it's better now.

andywer: Let's discuss this face-to-face next week. It can wait until then 🙂

ebma: I retried passing an `inputRef` to `<AssetSelector>` instead of using `<Controller>` but I still didn't manage to get it working.
What I did is:
- I wrap the AssetSelector with React.forwardRef 
- I add `inputRef={ref}` to the TextField of the AssetSelector
- I pass name="secondaryAsset" and `ref={form.register}` to the AssetSelector

But
Although `secondaryAsset` has a value at first (even in the props of the AssetSelector component on first render) it becomes undefined afterwards.
I tried adding the inputRef to the `SelectProps` and `InputProps` of the TextField but it didn't work either.

I can share my WIP code with you @andywer but right now I have no clue how to fix this.

andywer: It's fine – thx for looking into it again!
After reading the [Controller docs](https://react-hook-form.com/api#Controller) I think we might really need it.

PR #1016, opened 2020-03-20
Re-organize component source files

andywer:
Consistent source file re-organisation: No more source files directly in src/components/, but only in src/components/*


andywer:
Changes of this PR are included in #1033.

PR #1024, opened 2020-03-20
Update account-settings.json

Berniebaby:
testing testing

PR #1026, opened 2020-03-20
Update operations.json

Berniebaby:


andywer:
See https://github.com/satoshipay/solar/pull/1024#issuecomment-601830423.

PR #1033, opened 2020-03-24
Re-organize src/ directory

andywer:
Follow-up of #1016. This one is bold and subject to discussion.

The basic idea is to make the code base more modular and re-group the source files semantically like `src/App/`, `src/Account/`, …. You can import from other "modules" in src/ by a handy alias: `import … from "~Account/some/file"`, to import `src/Account/some/file`, for instance.

Common sub directories in `src/SomeModule`:
* `components`
* `hooks`
* `stories`

The goal is to re-organize the code in a way that it's more cohesive and feels more modular, thus hopefully reducing the number of merge conflicts in the future, esp. regarding potential forks of the app.


src/Account/components/AccountTransactions.tsx
```diff
@@ -3,17 +3,21 @@ import { useTranslation } from "react-i18next"
...
+import { hasSigned } from "~Generic/lib/transaction"
+import { useHorizonURL } from "~Generic/hooks/stellar"
+import {
+  useLiveRecentTransactions,
+  useLiveAccountData,
+  useOlderTransactions
+} from "~Generic/hooks/stellar-subscriptions"
+import { useIsMobile, useRouter } from "~Generic/hooks/userinterface"
+import { useLoadingState } from "~Generic/hooks/util"
 import * as routes from "../../routes"
```
ebma: Is there a reason why we don't do this for routes? We could just use `import * as routes from "~/routes"` everywhere right?

src/Account/components/AccountHeaderCard.tsx
```diff
@@ -4,16 +4,16 @@ import CardContent from "@material-ui/core/CardContent"
...
+import { useIsMobile, useRouter } from "~Generic/hooks/userinterface"
+import { matchesRoute } from "~Generic/lib/routes"
 import * as routes from "../../routes"
-import { breakpoints } from "../../theme"
-import { AccountCreation } from "../AccountCreation/types"
-import { Box } from "../Layout/Box"
-import withFallback from "../Lazy/withFallback"
-import ViewLoading from "../ViewLoading"
+import { breakpoints } from "~App/theme"
+import { AccountCreation } from "~AccountCreation/components/types"
```
ebma: Maybe we should put the `types.ts` file inside `src/AccountCreation/` rather than `src/AccountCreation/components/`?

(I'd rather have commented on the actual `src/AccountCreation/components/types.ts` file but couldn't because it was only renamed and not changed.)

src/Account/components/AccountTransactions.tsx
andywer: Yep, should work. The only reason is really that I used some regex search & replace and it wasn't matched by the expression 😉

src/Account/components/AccountHeaderCard.tsx
andywer: Good point!

PR #1036, opened 2020-03-26
Fix i18n key of CustomError

ebma:
The key for translated error messages is not correct anymore since we put the errors into the `generic.json` file.


andywer:
Great catch!

PR #1062, opened 2020-04-16
Add support for ledger hardware wallets

### LINKED ISSUE #686: Support Ledger hardware wallets
Ledger wallets come with a pretty convenient SDK: <https://github.com/LedgerHQ/ledgerjs>.

- Check for connected hardware wallets and show them in the accounts list
- Show some "hardware wallet" icon next to those accounts
- Abstractify account instances a bit, so that they work with hardware wallets, too

Open questions: Check out the SDK and see what needs to be changed technically.

ebma:
- [x] Add ledger-sdk dependencies + types
- [x] Add helper functions for using the ledger sdk
- [x] Add IPC for requesting hardware wallet accounts from within application
- [x] Add IPC for signing a transaction with a hardware wallet account
- [x] Add `isHardwareWalletAccount` flag to account instances 
- [x] Disable renaming and secret key export for hardware wallet accounts
- [x] Show adjusted version of `AccountDeletionDialog` to hardware wallet accounts that only enables users to merge accounts instead of deleting them
- [ ] Test with a physical device

Closes #686.


andywer:
Looks cool already! I think we should handle device detection slightly differently, though, otherwise we cannot handle hot-plugging.

The current implementation just checks if there is a device and keypair available when the React app code asks for it – which happens during initialization.

We should instead get the state once initially and then use the transport's `listen()` method to subscribe to device events: https://github.com/LedgerHQ/ledgerjs/blob/7d2f04f2f3ffc3eb8e8399c8d970569e4fa38326/packages/hw-transport/src/Transport.js#L34

When we receive a device add/remove event, we check if there is a keypair yet in case of a new device, then forward the information that a hardware wallet account just became available/unavailable to the React app.

Open question: Can we also subscribe to keypair added/removed events?

ebma:
I added your suggested changes. This also adds support for multiple connected ledger wallets which wouldn't have worked before.
I did not find anything related to keypair events but maybe it's not super bad if the user cannot add/remove them while the wallet is attached.

andywer:
Started testing with an actual Ledger Nano S yesterday – didn't work yet. Trouble with libusb, too, plus node-hid. We might also be using the Ledger SDK wrong.

I will try running everything in the web context…

ebma:
After being unsuccessful with connecting to the Ledger Nano X on Linux and Windows I dove deep into the code of the `@abandonware/noble` and `@ledgerjs/hw-transport-node-ble` package and was only partially successful. 
It took me a long time to figure out why the device was not found in the bluetooth discovery. You have to:
- (linux only) Run `sudo set-up cap_net_raw+eip ./node_modules/electron/dist/electron` because the local electron executable needs network privileges to run the bluetooth discovery
- (windows only) Use a compatible bluetooth 4.0 adapter and configure it for WinUSB (see [this](https://github.com/abandonware/node-bluetooth-hci-socket#windows))
- Change the parameter passed to startScanning() [here](https://github.com/LedgerHQ/ledgerjs/blob/96306b2c0d75e1290461fb52b8f69f506a425643/packages/hw-transport-node-ble/src/platform.js#L100) to [this](https://github.com/LedgerHQ/ledgerjs/blob/96306b2c0d75e1290461fb52b8f69f506a425643/packages/devices/src/index.js#L60)  but without the dashes (=>  `["13d634002c97000400024c6564676572"])`
It seems like the uuid of the services propagated by the ledger device is different for mac bindings compared to windows/linux because on mac it works without changing anything
-  Make sure the Ledger device is not already connected before running the bluetooth discovery


Besides doing all of this I still was not able to properly connect to the wallet device. When opening the connection the state changed to `connecting` but the connection never established properly (it remained `connecting`).

TL;DR: I tried tons of stuff and still didn't manage to get it working on windows/linux + there is an issue in the ledger package

While this is kind of a huge bummer I'm not sure if we wouldn't want to use it anyways. At least it's working perfectly fine on macOS and maybe we can fix the other issues in the future (or rather they get fixed by someone else).

ebma:
I re-added the `node-hid` dependency (`node-hid-singleton` to be exact because of issues with events on win10 probably similar to [this](https://github.com/LedgerHQ/ledgerjs/issues/430)) so that connecting hardware wallets via usb is now possible. I tested it successfully on macOS, win10 and (arch) linux. The bluetooth button + onboarding will only be visible on mac as this is the only platform where a bluetooth connection could be established (due to issues with pairing the device).

Building 64 bit windows executable is necessary because the `node-hid` dependency does not provide prebuilt binaries for win32 (see [here](https://www.npmjs.com/package/node-hid#supported-platforms)).

Windows users need to use [Zadig](https://zadig.akeo.ie/) to install a WinUSB driver for the ledger device (requirement of the [usb](https://www.npmjs.com/package/usb#windows) dependency).

andywer:
One more thing: Can we link to some documentation on how to get the USB driver to work on Windows from within the app if it's running on Windows? That would probably save us a bunch of support emails 😉

PS: Is the hardware wallet icon button hidden on mobile devices?

ebma:
I added an onboarding dialog which will open when you click on the icon next to the settings gear similar to what we have for bluetooth (but shown on windows only obviously).
Feel free to suggest wording improvements.
Screenshot 2020-08-17 at 16 43 11

And I'm not sure what you mean with 'hardware wallet icon button'. Do you mean the icon for hardware wallet accounts that is shown next to the account name similar to password protection?

PR #1079, opened 2020-05-11
Take open offers into account for trading

### LINKED ISSUE #1070: Take open offers into account for trading
Similar to #622 we should check the account liabilities and adjust the maximum amount that can be specified when trading.
This prevents the user from receiving "Not enough funds to perform this operation" errors.

ebma:
- [x] Pass amount of open orders to `getAccountMinimumBalance()` for calculation of trading form values so that open orders are considered when calculating minimum balance
- [x] Use `getSpendableBalance()` function which takes `selling_liabilities` into account
- [x] Add `getSpendableBalanceWithoutBaseReserve()` function which subtracts the base-reserve from the spendable balance if it's of type 'native' because issuing an order increases the minimum account balance
- [x] Show error message on submit when user cannot place new orders due to insufficient native balance

Closes #1070.

### DIFF
--- a/src/Trading/components/TradingForm.tsx
+++ b/src/Trading/components/TradingForm.tsx
@@ -25,7 +25,12 @@ import { useLiveOrderbook, useLiveAccountOffers } from "~Generic/hooks/stellar-s
 import { useIsMobile, RefStateObject } from "~Generic/hooks/userinterface"
 import { AccountData } from "~Generic/lib/account"
 import { CustomError } from "~Generic/lib/errors"
-import { balancelineToAsset } from "~Generic/lib/stellar"
+import {
+  balancelineToAsset,
+  getSpendableBalance,
+  getAccountMinimumBalance,
+  findMatchingBalanceLine
+} from "~Generic/lib/stellar"
 import { createTransaction } from "~Generic/lib/transaction"
 import { Box, HorizontalLayout, VerticalLayout } from "~Layout/components/Box"
 import { bigNumberToInputValue, isValidAmount, useCalculation, TradingFormValues } from "../hooks/form"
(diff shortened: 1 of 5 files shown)


src/Trading/hooks/form.ts
```diff
@@ -15,6 +15,18 @@ function findMatchingBalance(balances: AccountData["balances"], asset: Asset) {
   return balances.find(balance => balancelineToAsset(balance).equals(asset))
 }
 
+const baseReserve = BigNumber(0.5)
```
andywer, line 18: Please unify with https://github.com/satoshipay/solar/blob/a97c1b03c85b6446391d5dc7a8355103072e6657/src/Generic/lib/stellar.ts#L123-L124 😉

src/Trading/components/TradingForm.tsx
```diff
@@ -94,10 +99,18 @@ function TradingForm(props: Props) {
 
   const horizon = useHorizon(props.account.testnet)
   const tradePair = useLiveOrderbook(primaryAsset || Asset.native(), secondaryAsset, props.account.testnet)
+  const openOrders = useLiveAccountOffers(props.account.publicKey, props.account.testnet)
 
   const assets = React.useMemo(() => props.trustlines.map(balancelineToAsset), [props.trustlines])
 
-  const calculation = useCalculation(form.getValues(), tradePair, priceMode, props.accountData, props.primaryAction)
+  const calculation = useCalculation(
```
andywer, line 106: Now we are reaching the point where we have so many params that we should pass an options object and not single arguments anymore, so we can name them.

I can live with it for now, though. Just as a rule of thumb I think it's fair to say that starting with maybe 5 parameters you should consider an object and having six parameters you pretty much have to refactor if you don't wanna go crazy 😉

PR #1100, opened 2020-06-16
Fix fingerprint dialog called twice

### LINKED ISSUE #1077: Android: enabling fingerprint lock fails w/ BIOMETRIC_DISMISSED
Solar app installed from Google Play store; v0.24.0 on a Pixel 3a Android 10 build QQ2A.200405.005

When I attempt to enable Fingerprint lock, the app pops up the same "Confirm" dialog twice in succession, followed by two fingerprint unlocks, and results in the Fingerprint lock being disabled and a red "BIOMETRIC_DISMISSED" message at the bottom of the app.

Repro:
- Install the app from the Play store
- open the app (no accounts configured)
- open settings (gear icon)
- toggle Fingerprint lock
- dialog pops up: Confirm; Unlock your device once to enable the feature. Cancel OK
- click OK
- same dialog pops back up
- in quick succession: a Android Unlock Solar fingerprint dialog pops up, goes away, the BIOMETRIC_DISMISSED message flashes, and a second Android Unlock dialog appears
- doesn't matter whether you apply the fingerprint or cancel, you're returned to the settings page, Fingerprint lock is disabled

Repeatable: 100%

ebma:
Only use `props.onToggle` as `onClick` handler for the whole item instead of using it for item and `<SettingsToggle>`. Otherwise it would be called twice when the toggle is clicked.
(`onChange` is a required prop of `<SettingsToggle>` so `() => undefined` is passed instead)

Reject the bio-auth test with a `CustomError` to prevent an uncaught promise being thrown by `getErrorTranslation()`.

Closes #1077.

### DIFF
--- a/src/AppSettings/components/Settings.tsx
+++ b/src/AppSettings/components/Settings.tsx
@@ -64,6 +64,7 @@ export const BiometricLockSetting = React.memo(function BiometricLockSetting(pro
   return (
     <AppSettingsItem
       actions={
+        // pass empty onChange handler to prevent calling onToggle twice
         <SettingsToggle checked={props.enrolled && props.value} disabled={!props.enrolled} onChange={() => undefined} />
       }
       icon={<FingerprintIcon className={classes.icon} />}
--- a/src/AppSettings/components/Settings.tsx
+++ b/src/AppSettings/components/Settings.tsx
@@ -64,7 +64,7 @@ export const BiometricLockSetting = React.memo(function BiometricLockSetting(pro
   return (
     <AppSettingsItem
       actions={
-        <SettingsToggle checked={props.enrolled && props.value} disabled={!props.enrolled} onChange={props.onToggle} />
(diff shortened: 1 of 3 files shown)


src/AppSettings/components/Settings.tsx
```diff
@@ -64,7 +64,7 @@ export const BiometricLockSetting = React.memo(function BiometricLockSetting(pro
   return (
     <AppSettingsItem
       actions={
-        <SettingsToggle checked={props.enrolled && props.value} disabled={!props.enrolled} onChange={props.onToggle} />
+        <SettingsToggle checked={props.enrolled && props.value} disabled={!props.enrolled} onChange={() => undefined} />
```
andywer, line 67: Let's add a comment here why we set a do-nothing click handler here. LGTM afterwards 👍

ebma: Done

PR #1120, opened 2020-07-27
Fetch account data in horizon failover check

### LINKED ISSUE #1119: Try fetching account metadata as horizon health check
Right now we are only checking if the index route of the horizon server can be reached. That worked well so far, but the recent issues we had with horizon only surface when accessing the database, so we need to query some dynamic data.

Fetch some well-known account's (some account that is very unlikely to disappear, possibly friendbot on testnet) data as a health check. Any 200 response and possibly even 404 can be considered successful.

Follow-up of #1117.

ebma:
Fetches account data to see if querying dynamic data is possible.
`GAIH3ULLFQ4DGSECF2AR555KZ4KNDGEKN4AFI4SU2M7B43MGK3QJZNSR` is the account ID of the testnet friendbot but fortunately this account exists on pubnet as well so both requests should return with status `200`.

Closes #1119.

### DIFF
--- a/src/Workers/net-worker/stellar-network.ts
+++ b/src/Workers/net-worker/stellar-network.ts
@@ -94,7 +94,13 @@ export async function checkHorizonOrFailover(primaryHorizonURL: string, secondar
 
   try {
     // fetch dynamic data to check database access
-    const primaryResponse = await fetch(new URL(`/accounts/${testAccountID}`, primaryHorizonURL).href)
+    const primaryResponse = await Promise.race([
+      fetch(new URL(`/accounts/${testAccountID}`, primaryHorizonURL).href),
+      delay(2500).then(() => {
+        throw Error(`Horizon health check timed out. Trying failover…`)
+      })
+    ])
+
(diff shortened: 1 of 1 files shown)


src/Workers/net-worker/stellar-network.ts
```diff
@@ -92,7 +92,14 @@ export async function checkHorizonOrFailover(primaryHorizonURL: string, secondar
   try {
     const primaryResponse = await fetch(primaryHorizonURL)
     if (primaryResponse.ok) {
-      return primaryHorizonURL
+      // fetch dynamic data to check database access
+      const accountResponse = await fetch(
```
andywer, line 96: Two questions:
1. Why fetch the index route first and only then fetch the account? Doesn't seem to add value to the health check, but will now take twice as long until the health check returns a result.
2. What account is GAIH… (comment would be helpful 😉)? And will it likely always be available on both pubnet and testnet?

ebma: You're right, I somehow didn't realise the redundancy of both fetches. I changed it and added another comment that indicates that the account id belongs to the friendbot.

PR #1121, opened 2020-07-28
Make account settings accessible when offline

### LINKED ISSUE #1118: Allow user to export secret key even if account cannot be fetched
Cannot export secret key if account does not load since the account settings icon button is not shown. Make the account settings accessible even if offline.

Bildschirmfoto 2020-07-27 um 10 14 29

ebma:
Make account settings that are not relying on the fetched account data (i.e. exporting secret key, creating password, renaming account) accessible even when offline.

Closes #1118.


andywer:
Merge conflicts 😉

PR #1126, opened 2020-08-04
Use carousel in AccountSettings

### LINKED ISSUE #1009: Straighten account settings UI
There should be a carousel that transitions from the account settings list to a particular setting's dialog (like showing the change-password inputs).

Right now it will open a dedicated dialog for each setting which is pretty ugly and doesn't feel right from a user's perspective.

ebma:
- [x] Use a carousel instead of fullscreen dialogs for showing the account settings.
- [x] Hide back buttons of title components of account settings
- [x] Add `noMaxWidth` prop to `<DialogBody>` which disables the `maxWidth={900}` restriction (to allow increasing the width)
- [x] Set `noMaxWidth=true` when using `<DialogBody>` in account settings

Closes #1009.

### DIFF
--- a/src/Account/components/AccountView.tsx
+++ b/src/Account/components/AccountView.tsx
@@ -213,7 +213,11 @@ const AccountPageContent = React.memo(function AccountPageContent(props: Account
 
   const handleBackNavigation = React.useCallback(() => {
     if (props.account && matchesRoute(router.location.pathname, routes.accountSettings(props.account.id))) {
-      router.history.push(routes.account(props.account.id))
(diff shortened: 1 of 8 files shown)


andywer:
Great! We need to adapt the UI a little more, though.

1. Let's drop the back button in all settings. Let's make the back button in the blue header on top route back to the account settings main page instead.
2. Let's make the settings contents wider, so they are just as wide as the header content, the transaction history, …

ebma:
@andywer I added your suggested changes, please have a look again

Issue #1051, opened 2020-04-07
Error "destination is invalid" for anchorusd.com


andywer:
Happens on payment transaction creation when trying to withdraw from anchorusd.com.

andywer:
The error seems to be on anchorusd.com's side: When querying the transfer transaction from them, they return mostly `null` values.

![image](https://user-images.githubusercontent.com/1842462/78670113-93bd8c80-78dd-11ea-89b9-ad405d8332da.png)

We could still fail more explicitly, though.

PR #1133, opened 2020-08-11
Improve 'destination is invalid' error handling in withdrawal

ebma:
Throws a custom error with a more detailed explanation of why the transaction creation fails.
Catches the error so that the application does not crash and shows the message in an error notification instead.

Related to #1051 but does not actually fix the issue because it's caused by missing information of the anchor.


andywer:
> Change wording for i18n error as well

@ebma Thanks! 😉

PR #1138, opened 2020-08-24
Show federation record memo in send payment view

### LINKED ISSUE #714: Show federation record memo in "send payment" view
Whenever a user sets the destination to a federation address and that address' federation record demands setting a certain memo, we do enforce this memo and it's shown in the transaction summary dialog, but it is not visible before in the "send payment" view.

Furthermore if the user chose a custom memo, it will be silently replaced with the federation memo.

Background: We currently fetch the federation record when the user clicks on the "send" button. So we don't have the federation record yet at the time when the user fills out the form.

ebma:
- [x] Lookup federation record on destination input change
- [x] If a federated address enforces a memo disable the memo input field, fill the memo input with the required memo text and show `Memo (Text)` or `Memo (Id)` depending on type in the label
- [x] Show "Stellar Address does not exist" error message when destination matches scheme of a federated address but no record was found

Closes #714.

### DIFF
--- a/i18n/locales/en/payment.json
+++ b/i18n/locales/en/payment.json
@@ -32,6 +32,7 @@
     "not-enough-funds": "Not enough funds.",
     "invalid-destination": "Expected a public key or stellar address.",
     "invalid-price": "Invalid amount.",
-    "integer-memo-required": "Memo must be an integer."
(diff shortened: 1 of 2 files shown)


andywer:
Memo to myself: Still need to test this.

PR #1144, opened 2020-09-14
Fix android view not resizing with keyboard

### LINKED ISSUE #1000: Cannot scroll down in views on Android
Same as #960 but on Android.

There is a cordova preference which is useful in our case:
```
<preference name="resizeOnFullScreen" value="true" />
```
The [doc](https://github.com/ionic-team/cordova-plugin-ionic-keyboard#resizeonfullscreen-for-android-only) states that:
'There is an Android bug that prevents the keyboard from resizing the WebView when the app is in full screen (i.e. if StatusBar plugin is used to hide the StatusBar). This setting, if set to true, add a workaround that resizes the WebView even when the app is in full screen.' 
(Our app does not use the statusbar plugin but is running in full screen mode).

Unfortunately using this preference results in a black bar above the keyboard (see [here](https://github.com/ionic-team/cordova-plugin-ionic-keyboard/issues/132)) + the actions buttons on the bottom of the screen are slightly cut off. 

There seems to be a [working fix](https://github.com/ionic-team/cordova-plugin-ionic-keyboard/pull/134) for the black bar issue which is not merged yet.

So I'd suggest we wait until it is merged and add the preference afterwards or what would you think @andywer?

ebma:
Fixes an issue where focused input fields were covered by the keyboard if they are in the middle/bottom end of the screen and it's not possible to scroll down to make it visible again.
- [x] Set `resizeOnFullScreen` to `true` to make scrolling below the input fields possible
- [x] Add [this](https://github.com/ionic-team/cordova-plugin-ionic-keyboard/pull/145) fix for navigation overflow and the black bar issue (mentioned in #1000)
Since it's not yet merged into `master` and there are only changes in one file I decided to just hard-copy it to circumvent creating a new fork.

Closes #1000.

### DIFF
--- a/cordova/config/config.template
+++ b/cordova/config/config.template
@@ -117,6 +117,7 @@
     <preference name="HideKeyboardFormAccessoryBar" value="false" />
     <preference name="KeyboardResize" value="true" />
     <preference name="KeyboardResizeMode" value="native" />
+    <preference name="resizeOnFullScreen" value="true" />
     <preference name="ShowSplashScreenSpinner" value="false" />
     <preference name="SplashScreenDelay" value="$SPLASH_SCREEN_DELAY" />
     <preference name="SwiftVersion" value="5.0" />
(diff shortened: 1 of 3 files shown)


andywer:
Cool! Just one request:

Can you please add a comment to the top of the Java file, stating why you copied the file into the Solar repo (maybe with a link to this PR)? 😉

PR #1145, opened 2020-09-15
Adjust weightThreshold on signers change

### LINKED ISSUE #360: Auto-increase req. signatures when adding a co-signer
If the user does not want to change the required signatures, then they can always change it back again.

ebma:
Automatically adjust the shown amount of required signatures when user adds or removes signers in signer editor.

Closes #360.


andywer:
That was still implemented for the old multi-sig code.

PR #1146, opened 2020-09-21
Add button for printing secretkey backup sheet

ebma:
Adds a "Print" button to the secret-key export dialog. On 'initial-backup' the button is shown in the actions box next to the "Done" button and on default export it is shown next to the title of the inline dialog.

**on iOS:**
Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-21 at 16 56 06
Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-21 at 16 55 59
**on Android:**
Screenshot_1600703637
**in electron:**
Screenshot 2020-09-21 at 11 31 34


andywer:
Cool! Just one idea: What do you think about making the print landscape-oriented?

Secret Key Backup

(Not to be taken too seriously, I sketched this in 5mins 😉)

ebma:
I re-designed the layout of the backup sheet. It now looks like this:

Screenshot 2020-09-29 at 17 23 32
Screenshot 2020-09-29 at 16 52 14
Screenshot

Any suggestions for improvement of the bottom text are appreciated.

andywer:
@ebma How about this? I touch of color wouldn't hurt, but might look a little bit more balanced already.

Paper backup

ebma:
Hmm I guess it does not look so 'harsh' this way. I have to have a look if I can rotate 'Public Key' and 'Secret Key' to point upwards, IIRC I used some css attribute which worked better when rotating downwards like I did. Of course I can also do some fiddling with css-transform and stuff but do you have any hard feelings on the rotation here?

andywer:
No really hard feelings, just felt more natural to me that way.

Btw, wouldn't a simple `transform: rotate(270deg)` do the job? I could be wrong, though.

ebma:
I added your suggested changes. It now looks like this
Screenshot 2020-11-09 at 16 49 17
which should be pretty close to what you imagined.

PR #1147, opened 2020-09-22
Fix displaced layout of account settings on iOS/Android

ebma:
- [x] Fix actions in inline dialogs (Account Settings) not properly placed at the bottom of  the screen on iOS
- [x] Fix inline dialogs having too much spacing at the top on iOS and Android
This was due to the fact that the `<DialogBody>` components by default add additional spacing at the top on mobile to prevent the dialogs overlapping with notch spaces

**Before:**
Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-21 at 18 38 36
**After:**
Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-22 at 10 06 43

### DIFF
--- a/src/Layout/components/DialogBody.tsx
+++ b/src/Layout/components/DialogBody.tsx
@@ -55,7 +55,7 @@ interface Props {
   fitToShrink?: boolean
   noMaxWidth?: boolean
   preventActionsPlaceholder?: boolean
-  preventNotch?: boolean
+  preventNotchSpacing?: boolean
   top?: React.ReactNode
 }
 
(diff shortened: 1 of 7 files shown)


src/Layout/components/DialogBody.tsx
```diff
@@ -55,7 +55,7 @@ interface Props {
   fitToShrink?: boolean
   noMaxWidth?: boolean
   preventActionsPlaceholder?: boolean
-  preventNotch?: boolean
+  preventNotchSpacing?: boolean
```
andywer, line 58: 👍

PR #1150, opened 2020-10-05
Change style of network toggle button

### LINKED ISSUE #1148: Mobile network toggle is confusing
Should probably be a pill button or a dropdown rather than a toggle button as now it looks like a label indicating the network you're currently on.

So I'm no the testnet right?
IMG_3098

Wrong!
IMG_3099

ebma:
Changes the style of the network toggle button in the account list view to make it more obvious that it's a button and not a label indicating the current network.

- [x] Set a `border-radius` to make it look like a pill button
- [x] Add a `startIcon` on small screens
The icon is only shown on small screens because on larger screens the text `Switch to ...` makes it obvious that you are switching to that network.

Screenshot 2020-10-05 at 18 10 12
Screenshot 2020-10-05 at 18 03 15

No `startIcon` on large screens:
Screenshot 2020-10-05 at 18 03 39

Closes #1148.

### DIFF
--- a/src/App/theme.ts
+++ b/src/App/theme.ts
@@ -4,6 +4,7 @@ import Fade from "@material-ui/core/Fade"
 import ArrowDownIcon from "@material-ui/icons/KeyboardArrowDown"
 import amber from "@material-ui/core/colors/amber"
 import lightBlue from "@material-ui/core/colors/lightBlue"
+import grey from "@material-ui/core/colors/grey"
 import { SlideLeftTransition, SlideUpTransition } from "../Generic/components/Transitions"
 
 // TODO: The dark and light derivation of the brand color have not been design-reviewed!
@@ -276,6 +277,17 @@ const theme = createMuiTheme({
     MuiPaper: {
       rounded: {
         borderRadius: 8
+      }
+    },
+    MuiSwitch: {
+      colorSecondary: {
+        color: grey[50],
+        "&$checked": {
+          color: grey[50]
+        },
+        "&$checked + $track": {
+          backgroundColor: grey[50]
+        }
       }
     },
     MuiTab: {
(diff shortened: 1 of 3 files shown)


andywer:
Looks good, @ebma! I'm afraid we might not have solved the issue, though… 🙃

@ebma @tyvdh How about a [switch](https://material-ui.com/components/switches/#switch-with-formcontrollabel) instead? That should make it absolutely clear. On another note, if we were to use a switch component, we could use it as is for the desktop apps, too, I guess (-> no more dual behavior depending on platform).

ebma:
Hmm yeah I decided not to use a switch because I think it doesn't look that good. But you're right that it would make it absolutely clear. 
So how about a switch and to the right a "Testnet" label? Or no label at all? We probably don't have that much space on mobile.

andywer:
Didn't try, but I'd hope that there's enough space for a switch and a "Testnet" label.

I think visually it can look good if we just make the switch white. You can check if we are already using the primary/secondary variants of the switch somewhere. If not, it's probably a good idea to customize one of them (primary/secondary variant) to be white and use that.

ebma:
I replaced the button with a switch and it looks better than I expected.

Screenshot 2020-10-06 at 12 34 21
Screenshot 2020-10-06 at 12 34 37
Screenshot 2020-10-06 at 12 34 09

src/App/theme.ts
```diff
@@ -278,6 +279,17 @@ const theme = createMuiTheme({
         borderRadius: 8
       }
     },
+    MuiSwitch: {
```
andywer, line 282: You checked that the existing switches still look good? 😉

ebma: All other switches look the same since all of them have `color='primary'`. The one for toggling testnet is the only one that uses `color='secondary'`.

PR #1151, opened 2020-10-06
Unify components for account name and address

### LINKED ISSUE #277: Merge <AccountName> and <PublicKey>
There should only be one component for showing an account's name or public key.

ebma:
Group all components related to showing a public key or address under a new `<Address>` component which conditionally renders the best component based on props.

Closes #277.


src/Generic/components/Address.tsx
```diff
@@ -94,7 +96,7 @@ interface AddressProps {
 }
 
 // tslint:disable-next-line no-shadowed-variable
-export const Address = React.memo(function Address(props: AddressProps) {
+const BasicAddress = React.memo(function BasicAddress(props: BasicAddressProps) {
```
andywer, line 99: The name `BasicAddress` isn't great as it raises more questions than it answers. Tricky to find a better one, but how about `AddressContent`?

src/Account/components/TransactionList.tsx
```diff
@@ -106,13 +106,11 @@ function RemotePublicKeys(props: { publicKeys: string[]; short?: boolean; testne
...
   } else if (props.publicKeys.length === 1) {
-    return (
-      <PublicKey publicKey={props.publicKeys[0]} testnet={props.testnet} variant={props.short ? "short" : "full"} />
-    )
+    return <Address address={props.publicKeys[0]} testnet={props.testnet} variant={props.short ? "short" : "full"} />
   } else {
     return (
       <>
-        <PublicKey publicKey={props.publicKeys[0]} testnet={props.testnet} variant="short" />{" "}
+        <Address address={props.publicKeys[0]} testnet={props.testnet} />{" "}
```
andywer, line 113: Don't we still need to keep the `variant="short"`?

ebma: I decided to set `variant` to `"short"` by default because we almost always use `"short"` anyways.

src/Generic/components/Address.tsx
ebma: I agree that the name is bad. `AddressContent` sounds reasonable 👍

andywer:
@ebma This PR wants to be rebased 😉

PR #1152, opened 2020-10-06
Unify code for closing dialogs after transaction submission

### LINKED ISSUE #274: Unify code for closing dialogs after tx submission
Concerns all components using `<TransactionSender>`: There is no common logic when to close the dialog that uses `TransactionSender` after it successfully submitted a tx.

Current issue: Some components set a custom timeout to wait a bit for `TransactionSender`'s confirmation dialog to transition out before closing themselves, others don't.

ebma:
- [x] Add additional `completionCallbackDelay` prop to `<TransactionSender>` which controls the delay between registering the successful tx submission and triggering the `onSubmissionCompleted` callback. 
- [x] Remove custom timeouts for closing in components using the `<TransactionSender>`

Closes #274.

### DIFF
--- a/src/AccountSettings/components/AccountDeletionDialog.tsx
+++ b/src/AccountSettings/components/AccountDeletionDialog.tsx
@@ -92,7 +92,7 @@ interface AccountDeletionDialogProps {
   account: Account
   horizon: Server
   onClose: () => void
-  onDeleted: () => void
+  onDelete: () => void
   sendTransaction: (transaction: Transaction) => void
 }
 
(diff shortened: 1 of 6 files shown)


src/AccountSettings/components/AccountDeletionDialog.tsx
```diff
@@ -136,7 +130,6 @@ function AccountDeletionDialog(props: AccountDeletionDialogProps) {
       )
 
       await props.sendTransaction(transaction)
-      setTimeout(onDelete, 1000)
```
andywer, line 139: Doesn't this change mean that the dialog never gets closed in case of an account merge?

ebma: It will get closed because `onSubmissionCompleted` which is added on line 280 will call `onDelete` after the merge is done

andywer: Ahh, of course! I think I confused `props.onDelete` and `onDelete` for a moment.

Good that you cleaned that up! 👍

PR #1155, opened 2020-10-26
Fix error when adding asset to testnet account

ebma:
Fixes a bug where a `Cannot read property 'toLowerCase' of undefined` error is thrown when trying to add new assets to a testnet account.
The underlying problem is that the assets cache is not properly set up as it's sharing the same namespace as the well-known accounts cache.


andywer (approved):
👍

PR #1157, opened 2020-11-01
Add Italian translations

luca-90:
added  support for  italian language


src/App/i18n.ts
```diff
@@ -3,19 +3,23 @@ import LanguageDetector from "i18next-browser-languagedetector"
...
 import translationEN from "../../i18n/en"
+import translationIT from "../../i18n/it"
 
 i18n
   .use(LanguageDetector)
   .use(initReactI18next)
   .init({
     debug: process.env.NODE_ENV === "development" ? true : false,
-    fallbackLng: "en",
+    fallbackLng: "it",
```
andywer, line 13: Let's change it back to

```suggestion
    fallbackLng: "en",
```

and add `"it"` to `availableLanguages` in `i18n/index.ts`. Then the language selector will appear in the app settings, too 😉

andywer:
Thanks so much for this awesome contribution, @luca-90!

I left one comment, but haven't checked if the translations are complete (looks like a lot of content, though, so I guess it is) or correct (tricky, since I don't speak Italian 😉). Looks very good on first glance, though.

PS: Since you changed the fallback language to `it` – what's your operating system language? If it's Italian, the app should also automatically be in Italian then.

src/App/i18n.ts
luca-90: @andywer thank you very much for your review. As you suggested, I've just improved the translations for italian language, adding it into the availableLanguages and I've restored the default language to en.

andywer: Awesome, thank you so much! 🤩

@ebma Can you give it a try?

ebma: Sure!

luca-90: thank you so much @andywer . may ask you when do you think to release on app/play store the italian version of Solar wallet (If everything works fine of course) ? Also do you plan in the future to add the Stellar Federation server feature?

i18n/locales/it/account.json
```diff
@@ -0,0 +1,334 @@
+{
+  "action": {
+    "receive": "RICEVI",
+    "send": "INVIA"
+  },
+  "add-asset": {
+    "action": {
+      "remove": "Rimuovi",
```
ebma, line 8: Please add a translation for `"add-asset": "Add asset to account",` below line 7

ebma (changes requested):
Hey @luca-90, I reviewed your contribution and it's very good.
Most of my comments are only nit-picking over some minor things. 
Some remarks (on things that you could not know or are not your fault): 
- You should not change the words inside the double curly braces `{{..}}` because these will be replaced with values based on their name.
- When running your PR I noticed some strings that are still hard-coded thus I replaced them in another PR which is already merged into master. That's why I'm asking you in some comments to add another translation which was not there before. In order to do so you again need to merge the latest master into your branch.

i18n/locales/it/account.json
luca-90: thanks for the catch @ebma! I've just applied all the changes above in my latest commit. Sorry for changing values into curly braces {{..}} I didn't notice that

src/App/i18n.ts
andywer: We don't have a particular date in mind, but I guess it will be in the next few weeks as we've already lined up a bunch of smaller changes waiting to be released.

At the moment we don't plan to roll-out a federation server. We thought about it some while ago, but it hasn't really got any priority right now.

i18n/locales/it/account-settings.json
```diff
@@ -0,0 +1,221 @@
...
+        "item": {
+          "weight": "Peso",
+          "local-key": "Chiave locale"
+        }
+      },
+      "new-signer": {
+        "label": "Chiave Pubblica o indirizzo Stellar",
+        "placeholder": {
+          "long": "GABC… o indirizzo",
+          "short": "GABCDEFGHIJK… or alice*example.org"
```
ebma, line 117: One last thing: `"long"` and `"short"` need to be swapped and "or" should probably be "o", i.e.:
```
"long": "GABCDEFGHIJK… o alice*example.org", 
"short": "GABC… o indirizzo",
```

i18n/locales/it/account.json
ebma: Thanks! @andywer it might be a good idea if we added some guidelines on how to contribute new translations to the README and point out important things like 'Don't change values in curly braces, ...'?

andywer: @ebma Good idea! Can you prepare something? We should also state what files to touch and how to test it.

ebma: Yes I'll do it.

andywer:
@ebma Let's apply final adjustments after merging 😉

PR #1158, opened 2020-11-02
Change color of testnet UI

### LINKED ISSUE #1010: Highlight testnet UI more prominently
Maybe set a different background color or so. Let's make it more obvious that I as a user am in doing testnet stuff right now.

Then allow using duplicate account names on testnet and mainnet (allow accounts with same name on both networks, still disallow accounts with same name on same network).

ebma:
Changes the color of `primary` and `submit` buttons when using a testnet account and testnet badges to an orange color.
This way the user always has a more obvious indicator when doing things with a testnet account.

- [x] Add a `testnet` prop to the `<ActionButton>` component
- [x] Show the orange color for action buttons when the testnet prop is set and the button is of type `primary` or `submit`
- [x] Allow same account name for accounts on different networks

Closes #1010.


andywer:
@ebma One last thing: Let's make all testnet badges magenta then, incl. the one on the account view header 😉

ebma:
Done. Looks like this:
Screenshot 2020-11-11 at 11 16 38

By the way I noticed that the title of the trading dialog does not have a testnet badge next to it. Seems like a mistake since it's in most other relevant dialogs @andywer?

andywer:
@ebma Yes, let's add the badge to the trading dialog, too 👍

ebma:
I've added the testnet badge in a separate PR #1164.

PR #1159, opened 2020-11-04
Fix ambiguous error on request timeout

ebma:
- [x] Set `throwOnTimeout` option to `true` for the fetch queue because otherwise the promise will resolve `undefined` after the timeout is reached. 
- [x] Make `InlineErrorBoundary` translate the error like the `NotificationsContainer` and `MainErrorBoundary` do
- [x] Add a translation for `TimeoutError`

Setting `throwOnTimeout` will make `p-queue` throw a `new TimeoutError()` on timeout without any message provided. So the translation will make sure that we have an error message is shown.

Screenshot 2020-11-04 at 11 51 30


andywer:
Sounds good 👍

PR #1161, opened 2020-11-04
Enforce standard fontsize on android

### LINKED ISSUE #866: Everything looks super small on Google Pixel 3a XL
Seems like the screen has a very high resolution, but for some reason the web view doesn't adapt to it.

ebma:
This change will set the text zoom to `100%` (which is the default value) on app start thus making sure that UI elements are not displaced due to a high font size set by the user in system settings.

Closes #866.


andywer (approved):
Nice 👍

PR #1168, opened 2020-11-16
Improve security of electron app

### LINKED ISSUE #441: Try Electronegativity to screen app
See <https://blog.doyensec.com/2019/01/24/electronegativity.html>. Thanks for sharing, @PhilippLgh 🙂

ebma:
- [x] Add additional webPreferences to new instances of `BrowserWindow`
 for more details see: [disableBlinkFeatures](https://github.com/doyensec/electronegativity/wiki/AUXCLICK_JS_CHECK), [contextIsolation](https://github.com/doyensec/electronegativity/wiki/CONTEXT_ISOLATION_JS_CHECK) and [enableRemoteModule](https://github.com/doyensec/electronegativity/wiki/REMOTE_MODULE_JS_CHECK)
- [x] Check permission requests
- [x] Limit navigation flows of untrusted origins

Closes #441.

### DIFF
--- a/electron/src/preload.ts
+++ b/electron/src/preload.ts
@@ -1,5 +1,5 @@
 // tslint:disable-next-line: no-var-requires
-const { ipcRenderer } = require("electron")
+const { contextBridge, ipcRenderer } = require("electron")
 const electronProcess = process
 
 function sendMessage<Message extends keyof IPC.MessageType>(
@@ -55,7 +55,7 @@ const electron: ElectronContext = {
   subscribeToIPCMessages
 }
 
-global.electron = window.electron = electron
+contextBridge.exposeInMainWorld("electron", electron)
 
 process.once("loaded", () => {
   const newProcess = {
@@ -64,5 +64,5 @@ process.once("loaded", () => {
     platform: electronProcess.platform
   }
 
-  global.process = window.process = newProcess as NodeJS.Process
+  contextBridge.exposeInMainWorld("process", newProcess)
 })
--- a/electron/src/window.ts
+++ b/electron/src/window.ts
@@ -51,10 +51,27 @@ export function createMainWindow() {
 
   window.loadURL(webappURL)
 
+  window.webContents.session.setPermissionRequestHandler((webContents, permission, callback) => {
+    if (!webContents.getURL().startsWith("file://") && (permission === "media" || permission === "openExternal")) {
+      return callback(false)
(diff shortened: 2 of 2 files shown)


andywer:
Great stuff! Did you create a production build (can be unsigned) and test it yet?

electron/src/window.ts
```diff
@@ -48,10 +51,27 @@ export function createMainWindow() {
...
     event.preventDefault()
-    shell.openExternal(url)
+    if (window.webContents.getURL().startsWith("file://")) {
+      shell.openExternal(url)
+    }
+  })
+
+  window.webContents.on("will-navigate", event => {
+    // limit navigation flows to unstrusted origins
+    if (!window.webContents.getURL().startsWith("file://")) {
```
ebma, line 72: @andywer do you think this makes sense in our case? I'm not sure if I missed the point, the [example](https://github.com/doyensec/electronegativity/wiki/LIMIT_NAVIGATION_JS_CHECK) check is using a `https://...` url. 
So basically what I did should restrict the navigation to only `file://...` urls which should make sense because to open third-party URLs we always use `shell.openExternal`.

ebma:
> Great stuff! Did you create a production build (can be unsigned) and test it yet?

Good point because it actually did not work as it should. I was seeing accounts but not the ones stored on the file system but only the ones from local storage. The problem was related to setting `contextIsolation` to `true`.
We now have to use a so called `contextBridge` to expose objects from the `preload` script; there is an interesting [chapter](https://www.electronjs.org/docs/tutorial/context-isolation) in the documentation in case you want to know more about it.

electron/src/window.ts
andywer: Makes sense, but haven't tested it yet.

ebma:
I removed the two listener the two checks because I was not able to trigger the event listeners. [Here](https://www.electronjs.org/docs/api/web-contents#event-will-navigate) in the docs it's stated that the `will-navigate` event is not emitted when navigating programmatically which we always do with the `OpenLink` IPC message call.

electron/src/preload.ts
```diff
@@ -1,5 +1,5 @@
 // tslint:disable-next-line: no-var-requires
-const { ipcRenderer } = require("electron")
+const { contextBridge, ipcRenderer } = require("electron")
```
andywer, line 2: 👍

PR #1178, opened 2020-12-02
Update cordova dependencies

ebma:
- [x] Bump `cordova` to v10.0.0
- [x] Bump `cordova-android` to v9.0.0
- [x] Bump `cordova-ios` to v6.1.1
- [x] Remove `wkwebview-engine` plugin because it's now integrated in `cordova-ios` and the plugin is obsolete (more info [here](https://cordova.apache.org/announcements/2020/06/01/cordova-ios-release-6.0.0.html))
- [x] Remove `.env` file because it was only used to specify a custom `gradle-distribution-url` which is not needed anymore after bumping `cordova-android`

Some dependencies were automatically moved to `devDependencies` instead of `dependencies` by `cordova`.

### DIFF
--- a/cordova/.env
+++ b/cordova/.env
@@ -1,2 +1,0 @@
-# ATTENTION: THIS IS CHECKED INTO GIT
-CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL=https://services.gradle.org/distributions/gradle-4.10.1-all.zip
--- a/cordova/config/config.template
+++ b/cordova/config/config.template
@@ -96,12 +96,6 @@
         <edit-config target="NSCameraUsageDescription" file="*-Info.plist" mode="merge">
             <string>The app needs access to the camera, to be able to read QR codes.</string>
         </edit-config>
-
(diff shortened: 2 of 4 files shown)


andywer:
Any major changes we should be aware of? Or cool new features? 😉

ebma:
We would have run into problems some time ago if we had not used the `WKwebview` plugin before because it seems that Apple rejects apps that use UIWebView for some months now (see [here](https://cordova.apache.org/howto/2020/07/18/uiwebview-warning.html)). That's why they now use the `WKWebView` by default.
They bumped the minimum OS versions on Android from API level 21 to 22 and on iOS to 11.0.  
As far as I can tell most breaking changes are from bumping dependencies but I didn't see any that would cause issues for us.
Besides bug fixes and stuff they added more AndroidX and Kotlin support on android and on iOS they integrated some plugins into the `cordova-ios` package.
So nothing too special I'm afraid. 😅

ebma:
I might add that we need the updated `cordova-android` dependency because otherwise Google will reject our app since it does not target the latest sdk version.

PR #1179, opened 2020-12-02
Add support for SEP-0007 payment requests

### LINKED ISSUE #323: Support SEP-0007 payment requests
Support SEP-0007 `pay` operation. Only handle payment requests coming in via the signature coordination service if there is a system in place to defend against requests from untrusted parties.

ebma:
- [x] Add `transactionRequest` context (to parse and store incoming SEP-0007 requests)
- [x] Add `<TransactionRequestHandler>` that checks the `uri` stored in the `transactionRequest` context and shows dialogs according to its content
- [x] Add `<VerifyTrustedServiceDialog>` that is shown for incoming SEP-0007 requests that have an `origin_domain` but are not in the list of trusted services. The user can then decide to trust this domain and it will be added to the list of trusted services.
- [x] Add `<PaymentAccountSelectionDialog>` that is shown for incoming SEP-0007 `pay` requests. It shows details of the pay request and allows the user to select an account to use for this payment. Only accounts that hold a trustline for the specified asset will be shown in the list. A warning will be shown if the request is not signed. Selecting an account will show the payment dialog for the selected account.
- [x] Add `preselectedParams` as prop to the `<PaymentForm>` to be able to pre-fill the respective input fields with values of the `pay` request. Pre-filled input fields will be disabled so that the user cannot change the value. 
- [x] Enable trusted services by default (Remove the feature flag)
- [x] Save settings to `localStorage` in web build (makes it easier to test trusted services but is useful in general)

Closes #323.

**_Preview of the `<PaymentAccountSelectionDialog>` if no `origin_domain` and no `signature` are provided:_**
Screenshot 2020-12-02 at 15 22 31

**_Flow when `origin_domain` of request is already a trusted service:_**
pay-demo1

**_Flow when `origin_domain` of request is not a trusted service:_**
trusted-service-demo


i18n/locales/en/transaction-request.json
```diff
@@ -0,0 +1,38 @@
...
+    "footer": "Please select the account that you want to use for this payment.",
+    "uri-content": {
+      "any": "Any",
+      "pay": "Pay",
+      "to": "To",
+      "memo": "Memo",
+      "message": "Message"
+    },
+    "title": "Select Account for Payment",
+    "warning": "This request did not include a signature! Make sure that you trust the source of this link!"
```
andywer, line 24:
```suggestion
    "warning": "The origin of this request cannot be verified! Decline when in doubt."
```

andywer:
Great job, @ebma! 👍 

I proposed a wording change and we need to re-arrange the headlines in the view a little bit – looks somewhat unbalanced right now. Otherwise really cool.

We should also think about whether it makes sense to show the "trust this origin" dialog before the user has had a chance to see the actual request. Happy to fix this as a separate issue, though. Don't think it should be blocking.

src/Transaction/components/PaymentAccountSelectionDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+}))
+
+interface PaymentAccountSelectionDialogProps {
+  payStellarUri: PayStellarUri
+  onClose: () => void
+}
+
+function PaymentAccountSelectionDialog(props: PaymentAccountSelectionDialogProps) {
```
andywer, line 77: One thing that we can either refactor right now or we need to do it as soon as we start working on `tx` type requests:

It should probably rather be called `TransactionRequestReviewDialog` or similar and we should pass that SEP-7 URI content rendered as a grid as `children` prop, so we can re-use this component for the `tx` type URIs.

andywer:
Pushed one commit with a few UI adjustments. Btw, why did you prevent the max-width to be set? I thought it looked rather strange on a desktop screen. Maybe you had your reasons, but I changed that as part of my commit.

src/Transaction/components/PaymentAccountSelectionDialog.tsx
ebma: I suggest we postpone it until we do the `tx` type requests because it might be easier to restructure once we know how we handle/show `tx` requests.

ebma:
The `noMaxWidth` stems from the other dialogs that are used in the account settings. I did not pay attention on the max-width property when reusing code from those dialogs (for the sake of consistency) so it's a good thing that you noticed and changed it. 👍 
I'd also propose to tackle the issue with showing the 'trust this origin' dialog later.

ebma:
I changed it so that the `<PaymentForm>` is now shown inline instead of the user being redirected to the payment dialog. 

The payment form will not render without an account. That's why the first account in the account selection list is preselected by default. 
Screenshot 2021-01-20 at 16 08 22


If no accounts are available for the network specified in the tx-request another dialog is shown instead:
Screenshot 2021-01-20 at 15 49 54

ebma:
Superseded by #1202.

Issue #1177, opened 2020-11-30
Import of secret key does not work


AlessandroM92:
Hi folks,

in version 0.25.3 I cannot import my secret key. When I press the "Import Account" button nothing happens.

BR
Alessandro

ebma:
Hi @AlessandroM92, which platform are you using? I'm not able to reproduce this, is there any error shown above or below the textfields?

AlessandroM92:
I use the wallet on Windows 10. There are no errors shown.

ebma:
Please try opening the developer console by pressing `F12` or `CTRL+SHIFT+i` and see if clicking on the "Import Account" button will print an error message to the console. (Make sure the 'console' tab is selected after you opened the developer tools window).

AlessandroM92:
I actually get the following error: 

```
Uncaught Error: invalid checksum
    at k (bootstrap.541f4c3a.js:1011)
    at Function.value (bootstrap.541f4c3a.js:1011)
    at Function.value (bootstrap.541f4c3a.js:1014)
    at l (app-stage2.cec4b0ac.js:116)
    at f (app-stage2.cec4b0ac.js:116)
    at validateAccountCreation (app-stage2.cec4b0ac.js:116)
    at app-stage2.cec4b0ac.js:470
    at Object.l (bootstrap.541f4c3a.js:264)
    at s (bootstrap.541f4c3a.js:264)
    at f (bootstrap.541f4c3a.js:264)
```


```
k | @ | bootstrap.541f4c3a.js:1011
-- | -- | --
  | value | @ | bootstrap.541f4c3a.js:1011
  | value | @ | bootstrap.541f4c3a.js:1014
  | l | @ | app-stage2.cec4b0ac.js:116
  | f | @ | app-stage2.cec4b0ac.js:116
  | validateAccountCreation | @ | app-stage2.cec4b0ac.js:116
  | (anonymous) | @ | app-stage2.cec4b0ac.js:470
  | l | @ | bootstrap.541f4c3a.js:264
  | s | @ | bootstrap.541f4c3a.js:264
  | f | @ | bootstrap.541f4c3a.js:264
  | h | @ | bootstrap.541f4c3a.js:264
  | st | @ | bootstrap.541f4c3a.js:264
  | ut | @ | bootstrap.541f4c3a.js:264
  | ft | @ | bootstrap.541f4c3a.js:264
  | vt | @ | bootstrap.541f4c3a.js:264
[...]

PR #1182, opened 2020-12-07
Improve valid secret key check

ebma:
Use `Keypair.fromSecret()` instead of a regular expression to validate a secret key. This will do additional checks like validating the checksum. 

Doing so is necessary because there are cases in which a secret key matches the regular expression but is still invalid. 

Related to #1177.


andywer (approved):
Absolutely 👍

PR #1183, opened 2020-12-07
Fix hash memos not properly created in withdrawal

ebma:
Replaces the `createMemo` function declared in `useWithdrawalState.ts` with the `createMemo` function from `TransferService/util/utils.ts`. 
The function in `utils.ts` handles base64 decoding and hex encoding while the other does not.


andywer (approved):
👍

PR #1185, opened 2020-12-07
Improve items of language selector

### LINKED ISSUE #1176: Improve items of language selector
The language selector currently only shows abbreviations of languages. 
It would be better to show the whole language (i.e. `English`, `Italiano`, `Auto` instead of `en`, `it` and `auto`).

Nice to have: Instead of showing `Auto` as a separate item include it in the language that will be selected by default (i.e. `English (auto)` if english is the language that will be selected automatically).

ebma:
Changes the language selector so that it shows the full language name instead of its abbreviation. 
The language item that is equal to the browser language will show "(Auto)" at the end of its label.

Closes #1176.

### DIFF
--- a/i18n/index.ts
+++ b/i18n/index.ts
@@ -1,1 +1,6 @@
 export const availableLanguages = ["en", "it"] as const
+
+export const languageAbbreviationMap: { [type: string]: string } = {
+  ["en"]: "English",
+  ["it"]: "Italiano"
+}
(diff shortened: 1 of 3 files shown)


ebma:
I renamed it to `languageNames` and also added some left margin to the `Select` component because it was quite close to the text on smaller screens.

andywer (approved):

PR #1201, opened 2021-01-13
Fix valid balance check for deposits

ebma:
Fixes a bug where the user is not able to deposit funds if his balance is below the minimum amount required for withdrawals (which does not make sense but was an oversight).


andywer:
Good catch!

PR #1202, opened 2021-01-25
Add support for SEP-0007 pay and transaction requests

### LINKED ISSUE #323: Support SEP-0007 payment requests

ebma:
Supersedes #1179. 

- [x] Add `transactionRequest` context (to parse and store incoming SEP-0007 requests)
- [x] Add `<TransactionRequestHandler>` that checks the `uri` stored in the `transactionRequest` context and shows dialogs according to its content
- [x] Add `<VerifyTrustedServiceDialog>` that is shown for incoming SEP-0007 requests that have an `origin_domain` but are not in the list of trusted services. The user can then decide to trust this domain and it will be added to the list of trusted services.
- [x] Add `<PaymentRequestContent>` that renders an inline payment form and account selection list 
- [x] Add `<TransactionRequestContent>` that renders a transaction summary and an account selection list, either for selecting the source account (if a source account -replacement is defined) or for selecting the signing account
- [x] Add `<StellarRequestReviewDialog>` that will render the `PaymentRequestContent` or `TransactionRequestContent` based on the type of the incoming SEP-0007 request.
- [x] Add `preselectedParams` as a prop to the `<PaymentForm>` to be able to pre-fill the respective input fields with values of the `pay` request. Pre-filled input fields will be disabled so that the user cannot change the value. 
- [x] Enable trusted services by default (Remove the feature flag)
- [x] Save settings to `localStorage` in web build (makes it easier to test trusted services but is useful in general)
- [x] Add `<NoAccountsDialog>` that is shown if no accounts are available for the network specified in the SEP-0007 request

Closes #323.

Sample SEP-0007 requests can be found [here](https://lobstr.co/delegated-signing/test/) and [here](https://keybase.io/docs/cajaqx/lvcg/sep7links).

Payment request:
Screenshot 2021-01-20 at 16 08 22

Transaction request:
Screenshot 2021-01-25 at 12 05 10

No accounts:
Screenshot 2021-01-20 at 15 49 54


cblp:
Good work! We really need this in our DAO!

PR #1206, opened 2021-02-01
Fix app crashing due to unavailable transfer servers

ebma:
Fixes a bug where the application crashes due to network errors that occur when trying to deposit/withdraw with an account that trusts assets that do not have a transfer server.

Screenshot 2021-02-01 at 11 56 30

Catches the "no transfer server available" errors and returns an empty array.

### DIFF
--- a/src/Generic/hooks/transfer-server.ts
+++ b/src/Generic/hooks/transfer-server.ts
@@ -58,14 +58,15 @@ export function useTransferInfos(
     domain => {
       return (transferInfosCache.get(domain) ||
         transferInfosCache.suspend(domain, async () => {
-          try {
-            const transferServer = await initTransferServer(domain, testnet)
-            if (!transferServer) return []
+          const transferServer = await initTransferServer(domain, testnet)
+          if (!transferServer) return []
 
+          try {
             const transferInfo = await fetchTransferInfos(transferServer)
             return (transferInfo ? [transferInfo] : []) as [TransferServerInfo] | []
           } catch (error) {
-            // catch network error if no transfer server is available for domain
+            // tslint:disable-next-line no-console
+            console.error(`Failed to fetch transfer infos for transfer server ${transferServer.domain}:`, error)
             return []
           }
         }))[0]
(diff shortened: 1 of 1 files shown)


src/Generic/hooks/transfer-server.ts
```diff
@@ -58,11 +58,16 @@ export function useTransferInfos(
...
+          try {
+            const transferServer = await initTransferServer(domain, testnet)
+            if (!transferServer) return []
 
-          const transferInfo = await fetchTransferInfos(transferServer)
-          return (transferInfo ? [transferInfo] : []) as [TransferServerInfo] | []
+            const transferInfo = await fetchTransferInfos(transferServer)
+            return (transferInfo ? [transferInfo] : []) as [TransferServerInfo] | []
+          } catch (error) {
+            // catch network error if no transfer server is available for domain
```
andywer, line 68: Please log the error using `console.error()`, so that we know at least that an error occured 😉

src/Generic/hooks/transfer-server.ts
```diff
@@ -58,14 +58,15 @@ export function useTransferInfos(
     domain => {
       return (transferInfosCache.get(domain) ||
         transferInfosCache.suspend(domain, async () => {
-          try {
-            const transferServer = await initTransferServer(domain, testnet)
-            if (!transferServer) return []
+          const transferServer = await initTransferServer(domain, testnet)
```
andywer, line 61: Is it a good idea to move `initTransferServer()` out of the try-catch block? I think this call is just as likely to fail as the other one.

ebma: That's what I thought at first but it actually does its own error handling (see [here](https://github.com/satoshipay/solar/blob/f330a29fae737ba23c89eea61c5cedb0061b96f0/src/Generic/hooks/transfer-server.ts#L23)). So it'd be redundant to also include it in the try-catch block.

andywer: Good point, but i'd still keep it in the try-catch.

My thinking is that in this place in the code we don't know *how* `initTransferServer()` handles errors (unless you really look into the function's code you are calling). There is no guarantee that it will still catch any error tomorrow – maybe we change it to only catch certain errors that are anchor-related, for instance, and let others bubble up (like general errors that indicate an issue with our code). I doubt that whoever would change the `initTransferServer` function would also check every call to the function to make sure errors are handled properly by the caller.

WDYT?

ebma: I see your point and I guess you're right that it is more robust that way so I'll change it back.

PR #1207, opened 2021-02-01
Fix app crashing when searching for XLM in Add-Asset dialog

ebma:
Fixes a bug where the application crashes when entering "XLM" in the search bar of the add asset dialog.
If the item to be displayed is the native asset then `item.issuer` is "native" instead of a public key thus the `<AccountName>` component will try to fetch "native" which is invalid.


andywer:
No way! Stop finding those ridiculous issues 😅

No, but seriously… great catch! 👍

PR #1214, opened 2021-02-15
Add retry button to submission error view

### LINKED ISSUE #1208: Add "retry" button to submission error view
Allow easily retrying a failed submission.

PS: Set human-readable, translated error message on submission timeout error.

ebma:
- [x] Show a 'Retry' button next to the 'Close' button of the submission progress if the transaction submission fails
- [x] Show human-readable translated error message on submission timeout error

The issue of submission timeouts was not related to horizon but to the timeout specified for the promise queue [here](https://github.com/satoshipay/solar/blob/589241b2adfdae7d4fb2a8086de4420751e2c683/src/Workers/net-worker/stellar-network.ts#L101). That's why the transaction often went through although the "An unknown error occured" message was shown.
Instead of increasing this timeout I changed the error message shown in the submission progress so that the error translation is used if `error.response` is undefined (i.e. the error is not a submission error). 
This means that a "Request timed out" message (i18n translation for `timeout-error`) is shown once the timeout of 10 seconds is reached on transaction submission. If the user clicks on the newly added "Retry" button he is likely to see a "Successful" message pretty fast since the transaction was already accepted by the network.

Closes #1208.

### DIFF
--- a/i18n/locales/en/generic.json
+++ b/i18n/locales/en/generic.json
@@ -2,6 +2,9 @@
   "dialog-actions": {
     "close": {
       "label": "Close"
+    },
+    "retry": {
+      "label": "Retry"
     }
   },
   "error": {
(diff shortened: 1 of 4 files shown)


andywer (approved):
👍

PR #1216, opened 2021-02-15
Fix AppImage issues on linux

### LINKED ISSUE #1209: The SUID sandbox helper binary was found, but is not configured correctly
Failed to start app image for the first time as non-root user:
```
./Solar-Wallet-0.26.1.AppImage 
[24647:0202/140844.995112:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /tmp/.mount_Solar-TsNuku/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap (core dumped)
```

ebma:
Adds an "afterPack" script for  `electron-builder` that changes the permissions of the `chrome-sandbox` to `4755` to fix issues on some Linux distributions. This way users should not need to add the `--no-sandbox` attribute when running the `.appimage`.

I could not test it properly because I am not using a Debian based distribution.

Closes #1209.

### DIFF
--- a/electron-build.yml
+++ b/electron-build.yml
@@ -37,4 +37,5 @@ linux:
   artifactName: Solar-Wallet-${version}.${ext}
   category: Utility
 
+afterPack: ./scripts/electron-builder-afterpack.js
 afterSign: ./scripts/electron-builder-aftersign.js
--- a/scripts/electron-builder-afterpack.js
+++ b/scripts/electron-builder-afterpack.js
@@ -1,0 +1,31 @@
+const fs = require("fs")
+const path = require("path")
+
+const appDir = path.join(__dirname, "../electron/dist/linux-unpacked/")
+
+// Change permissions of the `chrome-sandbox` binary to 4755 to fix issues on some linux distributions
+module.exports = async function(context) {
+  const platform = context.packager.platform.name
+  // Only change permissions of `chrome-sandbox` on linux
+  if (platform !== "linux") {
+    return
+  }
+
+  console.log("Running linux afterpack hook...")
+  updateSandboxHelperPermissions(appDir)
+}
+
+async function sandboxHelperPath(appDir) {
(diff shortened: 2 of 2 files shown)


moranbw:
@ebma I tested this fix and I seem to get the following error:

```
The setuid sandbox is not running as root. Common causes:
  * An unprivileged process using ptrace on it, like a debugger.
  * A parent process set prctl(PR_SET_NO_NEW_PRIVS, ...)
Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
Trace/breakpoint trap
```
Do I need to clear any cache or anything like that?

- Debian 10 w/ MATE desktop

moranbw:
Doing some dig into Electron...it seems like this is a known issue of sorts. Not sure that 4755 will solve this.  

It seems like --no-sandbox is being suggested as the workaround/hack to get this to work. In some places (like snap) this has even been embedded into the `./AppRun`. To me this seems drastic -- first sentence in electron docs on sandbox: **"One of the key security features of Chromium is that all blink rendering/JavaScript code is executed within a sandbox."** https://www.electronjs.org/docs/api/sandbox-option

I'm still learning Electron, so I'm not an expert here.  But this seems like an issue worth addressing.

andywer:
@ebma Can you please respond to @moranbw? Must have slipped through.

andywer:
Nevermind. Just saw that the discussion was continued in the issue (#1209).

andywer:
@ebma @moranbw Can we act on this PR or should we close it for now?

ebma:
I guess we can close it for now. Although it fixes the original error about 'SUID sandbox helper binary was found, but is not configured correctly', it seems that there are more issues as @moranbw mentioned.

andywer:
On hold.

PR #1224, opened 2021-03-29
Disable and show loading indicator for submit button

### LINKED ISSUE #1221: "Place order" button not indicating progress
The "Place order" button is neither disabled nor showing progress with a spinner. This should be adjusted so that it behaves like the other submit buttons of the application.

ebma:
Closes #1221. 

@andywer I noticed that the submit buttons do not share the same behavior. For example, the "Send" button when creating a payment has a loading indicator but is not disabled (and I think if it's clicked again while loading, then the request will be sent again, which should not be possible).
Should I add a follow-up ticket about making all submit buttons behave similarly, i.e. showing a loading indicator + being disabled?


andywer:
@ebma Yes, a follow-up would make sense 👍

PR #1230, opened 2021-04-12
Improve invalid trading price handling

### LINKED ISSUE #1215: Trading error unclear if automatic price calculation fails
When trying to create a new order with the price set to 0.00 a "Couldn't find approximation" error is shown which does not really help the user identify the problem. This often happens when Solar is not able to automatically calculate a price for the asset pair. 

Expected: 
Highlight and expand the 'price' input field when clicking on "Place order" with a price of 0.00 and show an improved error message like 'Invalid price set'.

ebma:
Changes the validation rule for the trading price field so that 0 is not a valid input. The price input field is expanded if it has a validation error.

Closes #1215.

### DIFF
--- a/src/Trading/components/TradingForm.tsx
+++ b/src/Trading/components/TradingForm.tsx
@@ -348,7 +348,7 @@ function TradingForm(props: Props) {
         <ExpansionPanel
           className={classes.expansionPanel}
           elevation={0}
-          expanded={expanded || Boolean(form.errors.manualPrice)}
+          expanded={expanded}
           onChange={() => setExpanded(!expanded)}
         >
           <ExpansionPanelSummary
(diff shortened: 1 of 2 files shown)


src/Trading/components/TradingForm.tsx
```diff
@@ -375,7 +376,16 @@ function TradingForm(props: Props) {
               }
               control={form.control}
               name="manualPrice"
-              rules={{ validate: value => isValidAmount(value) || t<string>("trading.validation.invalid-price") }}
+              rules={{
+                validate: value => {
+                  const valid = isValidAmount(value) && !BigNumber(value).eq(0)
```
andywer, line 381: Does that maybe cause issues initially, before you enter any value?

ebma: The first validation is triggered after the user clicked on the "Place order" button and the `manualPrice` is set to 0 by default so there should not be any issues. If the user did not manually change the value and the calculation of an optimal price failed, the `manualPrice` will still be 0 and the validation rule will expand the field and return a proper error message.

andywer (approved):

PR #1231, opened 2021-04-12
Show badges for account import on wider screens

ebma:
At the moment, the badges for 'testnet' and 'password security' are only shown during account import for devices with small screens. It probably makes more sense to do it the other way round.

### DIFF
--- a/src/Account/components/AccountTitle.tsx
+++ b/src/Account/components/AccountTitle.tsx
@@ -310,7 +310,7 @@ function AccountTitle(props: AccountTitleProps) {
   return (
     <MainTitle
       actions={props.actions}
-      badges={props.editable && !isSmallScreen ? null : props.badges}
+      badges={props.editable && isSmallScreen ? null : props.badges}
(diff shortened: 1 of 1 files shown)


andywer:
No, it actually makes sense the way it is.

On wider screens the account name and the close and menu icon buttons are on the same line, so there is ironically less space. On mobile devices the icon buttons are on a row of their own, so there is enough space on the account name row while editing the account name.

Feel free, however, to add a comment explaining that 🙂

ebma:
While I kind of see your point I'd argue that it's different in practice.

On mobile, it's taking so much space that the default name appears with an ellipsis.
Screenshot 2021-04-26 at 12 51 02

This is at a width of 600px so at the magic break point where we do the differentiation.
Screenshot 2021-04-26 at 12 53 39

And this is at 601px.
Screenshot 2021-04-26 at 12 53 16

So I'd say there is enough space and the line break does not make much of a difference.

PR #1232, opened 2021-04-14
Fix app crashing due to web auth challenge error

ebma:
Fixes a bug where the application crashes in deposit/withdrawal if fetching the sep-10 web auth challenge fails.


andywer (approved):
Nice 👍

PR #1240, opened 2021-05-03
Fix Spanish language not configured properly and missing multi-sig translations

ebma:
The Spanish translations were not added to the i18n `resources` so they were not loaded when switching the language in the app settings.

I noticed that there are no Spanish translations for the new multi-sig strings, @gonzamontiel could you maybe add those as well? I already added placeholders to the JSON files so that you don't have to spend time figuring out which are missing.

### DIFF
--- a/i18n/locales/es/account-settings.json
+++ b/i18n/locales/es/account-settings.json
@@ -93,15 +93,40 @@
   },
   "manage-signers": {
     "action": {
-      "add-co-signer": {
-        "long": "Añadir co-firmante",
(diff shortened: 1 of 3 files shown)


gonzamontiel:
> I noticed that there are no Spanish translations for the new multi-sig strings, @gonzamontiel could you maybe add those as well? I already added placeholders to the JSON files so that you don't have to spend time figuring out which are missing.

Sure thing, on it.

ebma:
Nice, thank you very much!

PR #1241, opened 2021-05-06
Quick-fix for missing balance issue

ebma:
The original issue was a wrong `Location` parameter in the redirection response.
Screenshot 2021-05-06 at 11 34 03

Although we fixed this, the balance is still missing if the user does not re-install the application due to the long expiry time of the cached response.

The easiest way to circumvent this issue, for now, is to just use the SDF horizon.

### DIFF
--- a/src/App/contexts/stellar.tsx
+++ b/src/App/contexts/stellar.tsx
@@ -18,11 +18,8 @@ const initialHorizonSelection = (async () => {
   const { netWorker } = await workers
 
   return Promise.all([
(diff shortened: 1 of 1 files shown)


andywer:
@ebma Ha! This didn't even get merged… Should we close it?

ebma:
Did you include this in the updated version of v0.27.0? I used it in the updated android build. But we can close it without merging since it was only a short-term solution anyways.

andywer:
No, I did not. On iOS it worked without this quick-fix, so did the desktop builds. At least mac OS, but I didn't see any complaints from Windows or Linux users if I remember correctly, so I think it's safe to close.

PR #1248, opened 2021-05-12
Fix invalid price error in trading form

ebma:
Fixes a bug where the user cannot submit a trade order because the "Invalid price" error message is always shown if the user has not manually changed the price.

I removed the validation from the `<Controller>` component because the value that is passed here is not enough to check if a price is invalid. The reason for this is that the price input field does either contain the `manualPrice` (input by the user) or the `defaultPrice` that was calculated in the `useCalculation` hook. The `validation` function however receives only the `manualPrice` (that is 0 until the user edited the field manually) which is basically the root issue of the bug. I tried passing a validation function that uses the calculation results of the `useCalculation` hook but this does not work because the component is not rerendered and the function not up-to-date. That's why I put the validation into the `submitForm()` function.

### DIFF
--- a/src/Trading/components/TradingForm.tsx
+++ b/src/Trading/components/TradingForm.tsx
@@ -97,7 +97,7 @@ function TradingForm(props: Props) {
   })
 
   const sendTransaction = props.sendTransaction
-  const { primaryAsset, secondaryAsset } = form.watch()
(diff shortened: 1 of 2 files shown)


andywer:
@ebma Can we close #1251 now?

ebma:
@andy not quite sure, let's maybe wait for an answer from @marcoachury, he could have meant something different.

PR #1252, opened 2021-05-26
Add horizon scheduling mechanism

ebma:
Changes `useHorizonURL()` to return one of the multiple available horizon URLs in a round-robin-like manner.
This is supposed to help prevent timeouts of requests.


andywer (changes requested):
I'm afraid we cannot merge it like this. All caching and network deduplication depends on the horizon URL.

If we now return *some* horizon URL (non-deterministically from a caller's perspective) whenever a connection to horizon is necessary, these caches will not work properly anymore, resulting in even more unnecessary requests (tested it).

ebma:
I added some changes that fix the caching issue and improve the performance of requests a lot.

andywer (approved):
I tweaked the code a little bit more, but now the transaction history wouldn't load anymore, so I finally decided to drop my efforts, revert your last two commits and go with that…

PR #1253, opened 2021-05-31
Fix trading form input field issues

### LINKED ISSUE #1251: Problem entering price at: Trade / Sell / Advanced
I noted this on 0.27.0

Fail to validate keyboard entry on Trade/Sell/Advanced...  May be related with spanish locale, not clear if decimal separator is "." or ","

imagen

imagen

ebma:
Fixes issues on iOS where the user was unable to enter a valid amount because the on-screen keyboard only showed a comma which was not accepted by the validation mechanism (when the device is set up for EU regions).

Seems like this issue appears on iOS only. When entering a value with a comma into the input field (either the primary amount or manual price, does not matter) it is not propagated correctly, i.e. the onChange event only contains `""`. Removing the `type: "number"` prop helps here, but this is not enough as `Big.js` somehow still has issues with the comma string and throws an `Invalid number` error when trying to instantiate a new `Big()`. That's where the `replaceCommaWithDot(input:string)` function comes into play. 

Also closes #1251 by re-introducing the validation rule such that the price input field will already show an "Invalid amount" error instead of first showing it when the user clicks on the submit button.

### DIFF
--- a/src/Trading/components/TradingPrice.tsx
+++ b/src/Trading/components/TradingPrice.tsx
@@ -64,7 +64,6 @@ const TradingPrice = React.forwardRef(function TradingPrice(props: TradingPriceP
       onChange={props.onChange}
       onFocus={props.selectOnFocus ? event => event.target.select() : undefined}
       style={props.style}
-      type="number"
       value={props.defaultPrice ? props.defaultPrice : props.manualPrice}
     />
   )
--- a/src/Trading/components/TradingPrice.tsx
+++ b/src/Trading/components/TradingPrice.tsx
@@ -52,7 +52,7 @@ const TradingPrice = React.forwardRef(function TradingPrice(props: TradingPriceP
   return (
     <TextField
       inputProps={{
-        pattern: "[0-9]*",
+        pattern: "^[0-9]*(.[0-9]+)?$",
         inputMode: "decimal",
         min: "0.0000001"
       }}
--- a/src/Trading/hooks/form.ts
+++ b/src/Trading/hooks/form.ts
@@ -8,7 +8,7 @@ import { BASE_RESERVE, balancelineToAsset, getAccountMinimumBalance, getSpendabl
 import { useConversionOffers } from "./conversion"
 
 export const bigNumberToInputValue = (bignum: BigNumber, overrides?: BalanceFormattingOptions) =>
-  formatBalance(bignum, { minimumSignificants: 3, maximumSignificants: 9, ...overrides })
+  formatBalance(bignum, { minimumSignificants: 3, maximumSignificants: 9, groupThousands: false, ...overrides })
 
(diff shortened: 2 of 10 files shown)


src/Trading/components/TradingPrice.tsx
```diff
@@ -64,7 +64,6 @@ const TradingPrice = React.forwardRef(function TradingPrice(props: TradingPriceP
       onChange={props.onChange}
       onFocus={props.selectOnFocus ? event => event.target.select() : undefined}
       style={props.style}
-      type="number"
```
andywer, line 67: @ebma Do we still get to see the numpad rather than the default keyboard? 😉

src/Trading/hooks/form.ts
```diff
@@ -25,6 +25,12 @@ function getSpendableBalanceWithoutBaseReserve(accountMinimumBalance: BigNumber,
   return spendableBalance.cmp(BigNumber(0)) < 0 ? BigNumber(0) : spendableBalance
 }
 
+// replaces ',' with '.' in a string
+// this can be used with strings that represent a number before passing them to Big()
+export function replaceCommaWithDot(input: string) {
+  return input.replace(/,/g, ".")
```
andywer, line 31: Is there anything we can do to prevent accidents in the case that someone tries to enter 3000 as `3,000`?

Bonus question: Does the form do its job if I enter 0.25 as `.25`?

andywer: One more thing:
Now is logic is custom for the trading view. Actually, it should be consistent across all our forms… 🙈

src/Trading/components/TradingPrice.tsx
ebma: Yes, at least on iOS (didn't check on Android). I guess it's related to the pattern `pattern: "[0-9]*",` and not necessarily the `type="number"`.

src/Trading/hooks/form.ts
ebma: Regarding the bonus question: yes it would work if we change the regex pattern. Right now it will expect something like `x.xx` but seems like Big.js can also use strings without leading 0.
And I don't think we can do much about the `3,000` meaning 3000 or at least not without much effort. Seems like this locale thing is quite complex and to be honest I still don't fully understand why iOS is causing this issue anyways.

ebma: I added the additional handling to the payment form as well, and I think the trading and payment form would be the only places where we actually expect inputs with fractions? So it should be fine, but I see your point that we could spend more efforts to fully prevent this issue in the future.

ebma:
I moved the new utility functions to a new file and created a new function called `FormBigNumber()` which should be used instead of `BigNumber()` whenever we are instantiating a Big with user input. 
And I also changed the regex such that `.xx` or `,xx` are also accepted.

